PR #11, opened 2015-08-21
Fix python smart pointers

asylvest:
Mega test_create_sicd_xml.py program helped catch a lot of loose ends with the Python bindings.  The biggest problem was that the smart pointers needed functions in Swig to actually create them... @chvink suggested an approach in mem.i to use a macro to create all these.  This also highlighted a bunch of smart pointers that we had missed.


asylvest:
As part of this, several types changed... where possible, I removed nonsensical clone() methods.  There were a couple raw pointers that had still been lingering in C++ that got upgraded to smart pointers as part of this.

PR #12, opened 2015-08-22
Fix Windows Python test

asylvest:
If we don't close the FileOutputStream prior to re-opening it with a FileInputStream, we get a permissions error on Windows


asylvest:
@chvink make sure CPHD test does this close too

PR #18, opened 2015-09-17
New Projection and static utilities

bombaci-vsc:
Here are some changes which add a Contour to Image projection and refactors some existing projection functions to not require creating a projection model object

### DIFF
--- a/modules/c++/scene/include/scene/ProjectionModel.h
+++ b/modules/c++/scene/include/scene/ProjectionModel.h
@@ -95,6 +95,18 @@ public:
      *  R/Rdot Contour Ground Plane Intersection
      *
      */
+    static Vector3
+        contourToGroundPlane(double rCOA, double rDotCOA,
+                             const Vector3& arpCOA,
+                             const Vector3& velCOA,
+                             const Vector3& groundPlaneNormal,
+                             const Vector3& groundRefPoint,
+                             int look);
+    /*!
+     *  Calculations for section 5.2 in SICD Image Projections:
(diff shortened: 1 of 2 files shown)


asylvest:
Thanks.  Do you have an ETA for when the equations for the fullContourToImageCoord() will appear in an updated SICD Volume 3?  Is there a draft copy that you could send?

chvink:
looks like we'll want to regenerate scene's python bindings, too?

modules/c++/scene/include/scene/ProjectionModel.h
```diff
@@ -95,6 +95,18 @@ class ProjectionModel
      *  R/Rdot Contour Ground Plane Intersection
      *
      */
+    static Vector3
```
asylvest: I'm not necessarily opposed to having a static method like this, but what's the use case where you want to just get the contour but don't want to construct a ProjectionModel?

asylvest:
@chvink let's wait until the dust settles on the C++ side but yeah once that's finalized we'll want to re-generate the Python bindings.

modules/c++/scene/include/scene/ProjectionModel.h
mtrachy: The main use case I can think of is when doing projections prior to (or during) image formation, when the entirety of the information needed to construct a ProjectionModel is not available.

asylvest:
OBE - if this ever is something we want to move into SIX, we'll certainly want to revisit the implementation that was started here (hopefully with the corresponding equations documented in the SICD spec)

PR #20, opened 2015-10-05
Sicd utility functions

chvink:
The two functions now take a NITFReadControl instead of the paths, and there's extra functions that takes care of both for you.


modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -19,29 +19,140 @@
...
+     *
+     * \param sicdPathname SICD NITF pathname
+     * \param schemaPaths One or more files or directories containing SICD
+     * schemas
+     *
+     * \return a NITFReadControl 'load()'ed with the given SICD
+     *
+     */
+    static
+    NITFReadControl getLoadedNITFReadControl(const std::string& sicdPathname,
```
asylvest: NITFReadControl is not safe to copy.  To that end, make it noncopyable (private, unimplemented copy constructor and assignment operator).  I have to look at the rest of the new code still, but if this function sticks around, have it return an auto_ptr.

modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -19,29 +19,140 @@
...
+     * parses the SICD's XML and returns a new ComplexData object.
+     *
+     * \param reader A NITFReadControl loaded with the desired SICD
+     *
+     * \return ComplexData new ComplexData associated with the SICD NITF
+     *
+     * \throws except::Exception if the provided reader is not a SICD
+     *
+     */
+    static ComplexData* getComplexData(NITFReadControl& reader);
```
asylvest: Return auto_ptr

modules/c++/six.sicd/source/Utilities.cpp
```diff
@@ -130,6 +130,166 @@ Utilities::getProjectionModel(const ComplexData* data,
...
+        const std::string& sicdPathname,
+        const std::vector<std::string>& schemaPaths)
+{
+    six::XMLControlRegistry xmlRegistry;
+    xmlRegistry.addCreator(six::DataType::COMPLEX,
+            new six::XMLControlCreatorT<
+            six::sicd::ComplexXMLControl>());
+
+    six::NITFReadControl reader;
+    reader.setXMLControlRegistry(&xmlRegistry);
```
asylvest: Eh, is this safe?  The read control is holding onto a pointer to the registry, and is obviously not taking ownership since the registry is on the stack.  Is the registry only used during the load() when it parses the XML?  If someone tried to do another load() with the returned reader, they'd get a seg fault because they'd be holding onto a pointer to an invalid registry right?  Assuming this is all true, after the load(), add a setXMLControlRegistry(NULL) w/ a comment to make this safer.

Longer term I think we want to drop the registry altogether as it's not used like it was originally intended and just makes stuff a headache.

modules/c++/six.sicd/source/Utilities.cpp
```diff
@@ -130,6 +130,166 @@ Utilities::getProjectionModel(const ComplexData* data,
...
+    // as cloneable pointers - removeData() deletes the object)
+    // TODO: shared pointers
+    retv.complexData = reinterpret_cast<ComplexData*>(complexData->clone());
+    retv.widebandData = buffer;
+    return retv;
+}
+
+ComplexData* six::sicd::Utilities::getComplexData(NITFReadControl& reader)
+{
+    // TODO: generalize this to multi-image SICDs
```
asylvest: Remove TODO - there is no such thing as a multi-image SICD.  It will always be just one image (even if under the hood it is multiple NITF image segments)

modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -19,29 +19,140 @@
...
+     * \return a pointer to the loaded data.
+     *
+     * \throws except::Exception if the pixel type of the SICD is not a
+     *           complex float32 or complex int64, or
+     *         if the user provided buffer is not large enough to store
+     *           the whole image in, or
+     *         if a buffer size was provided but the buffer pointer is null
+     *
+     */
+    static std::complex<float>* getWidebandData(
```
asylvest: So I agree in principle that this is efficient but it's also easy to accidentally leak memory.  I think there should be two overloadings.  1: you pass in a pre-allocated buffer and it better be big enough to fit or 2: you pass in a `std::vector<std::complex<float>>` which gets resized for you.  The latter can turn around and call the former.  If there are any overloadings where you just pass in a filename (so the caller can't trivially know the size needed upfront since they haven't looked at the ComplexData yet), let's only have a vector version for now.

modules/c++/six.sicd/include/six/sicd/Utilities.h
chvink: Is there a reason to use auto_ptr over something from mem in coda-oss? I avoided auto_ptr purposefully because I didn't want to add more code that we know is deprecated in C++11.

modules/c++/six.sicd/source/Utilities.cpp
chvink: We could statically add all of Six's XMLControlCreator to the singleton XMLControlFactory, that way we'd never have to explicitly touch the XMLRegistry when we want to load a file. The weird thing (that is already there, I guess) is that six::Utilities::parseData() needs to know about SICD and SIDD, and I guess it just doesn't know about CPHD. We could get around that by adding a virtual `canParse(string)` function to each XMLControlCreator (or somewhere convenient). That way XMLControlRegistry can ask all of its creators if they know how to parse particular XML and return the right one without parseData knowing about what it's loading.

modules/c++/six.sicd/include/six/sicd/Utilities.h
asylvest: So here is my rationale on this.  On the happy day that we can move over to C++11, big chunks of mem will themselves be deprecated too (i.e. you could use a `mem::SharedPtr` now, but at the point that we're all-in with C++11, we'll remove it anyway since it's just a poor man's `std::shared_ptr`).  There are two schools of thought... Scott Meyers says to just always use shared pointers since they'll do what you want.  I prefer to clearly express the ownership based on the smart pointer type.  So, by returning an `auto_ptr`, you're making it clear that you're the only owner of this thing (which in this case as the caller you probably don't care but it still adds a little clarity).  When we go to C++11, we'll do a find/replace and change all `std::auto_ptr`s over to the new and improved `std::unique_ptr`.  In the meantime, I turned off the gcc warning (in build.py) about auto_ptr being deprecated since there's nothing we can do about it presently.

modules/c++/six.sicd/source/Utilities.cpp
asylvest: For now, I'd like to avoid using the singleton factory when possible (for potential thread safety reasons as well as my own pedantic hatred of singletons).  Originally the intent of the XML control stuff was to allow users to register their own XML controls, but in real life no one does this and as you mentioned, CPHD never dealt with any of this hoopla (and works just fine).  So, rather than going the canParse() route, I think we just live with the current XML control stuff for now, trying to avoid singleton use when practical, and eventually (New Year's resolution) cut it out altogether.

chvink:
These should all be addressed. `getLoadedNITFReadControl()` is gone now, auto_ptrs are in, vector `getWidebandData()` overload is in.

modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -19,30 +19,172 @@
...
+     * \return ComplexData associated with the SICD NITF,
+     *           allocated with new
+     *         std::complex<float> buffer containing SICD wideband data,
+     *           allocated with new[]
+     *
+     * \throws See six::sicd::Utilities::getComplexData and 
+     *           six::sicd::Utilities::getWidebandData
+     *
+     */
+    static SicdContents readSicd(const std::string& sicdPathname,
```
asylvest: Rather than the complexity of a new `SicdContents` class (and the copy that will occur upon returning this unless this is a C++11 compiler or it's smart enough to do return value optimization), just have this function take in an auto_ptr to ComplexData by ref and a vector of complex float by ref.

modules/c++/six.sicd/include/six/sicd/Utilities.h
asylvest: I'm still worried about accidentally leaking memory with this version, so I think this should be changed to be

```
static void getWidebandData(NITFReadControl& reader,
                            const ComplexData& complexData,
                            std::complex<float>* buffer);
```

modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -19,30 +19,172 @@
...
+     *
+     */
+    static void getWidebandData(
+            NITFReadControl& reader,
+            const ComplexData& complexData,
+            std::vector<std::complex<float> >& buffer
+            );
+
+    /*
+     * From here down is the legacy interface for
```
asylvest: I don't think you need this todo... I think keeping these old functions is fine

modules/c++/six.sicd/source/Utilities.cpp
asylvest: Still need to set the registry to NULL on the reader here

modules/c++/six.sicd/source/Utilities.cpp
```diff
@@ -130,6 +130,160 @@ Utilities::getProjectionModel(const ComplexData* data,
...
+        region.setStartRow(startRow);
+        region.setStartCol(startCol);
+        region.setNumRows(numRows);
+        region.setNumCols(numCols);
+        region.setBuffer(reinterpret_cast<UByte*>(returnBuffer));
+        reader.interleaved(region, imageNumber);
+    }
+    else if(pixelType == PixelType::RE16I_IM16I)
+    {
+        const size_t elementsPerRow = numCols*2;
```
asylvest: Is there a way to avoid this code duplication between the two functions where we do this reading N rows at a time and cast kind of stuff?

modules/c++/six.sicd/source/Utilities.cpp
```diff
@@ -130,6 +130,160 @@ Utilities::getProjectionModel(const ComplexData* data,
...
+
+    return returnBuffer;
+}
+
+void Utilities::getWidebandData(NITFReadControl& reader,
+                                const ComplexData& complexData,
+                                std::vector<std::complex<float> >& buffer)
+{
+    size_t requiredNumElements = complexData.getNumCols() * complexData.getNumRows();
+    buffer.resize(requiredNumElements);
```
asylvest: To be pedantic, need to handle the case where requiredNumElements is 0 in which case `&buffer[0]` is not valid to do.  Easiest thing is probably just to only call `getWidebandData()` if `!buffer.empty()`.  Also, don't bother with the extra parens - just use `&buffer[0]` rather than `&(buffer[0])`.

chvink:
Everything should be updated based on these comments.

asylvest:
For expediency I made a few more changes directly myself.  These were related to style, correcting a comment, and avoiding code duplication.

PR #29, opened 2015-11-06
Add legend reading support

asylvest:
Previously, `NITFReadControl` did not actually read the `Legend` into the `Container`.  Added support for this along with a unittest showing this all works properly.


asylvest:
@chvink

PR #31, opened 2015-11-20
BufferList should hold const pointers, not just pointers

asylvest:
@chvink


asylvest:
Changing typedef and then rippling through where the compiler can't auto-convert the type.

PR #32, opened 2015-11-24
Extend projection model python interface

dstarinshak:
Update extends the Python interface to ```ProjectionModel::imageToScene```. Currently, the member function may be called with a ```RowCol<double>``` type for (row, column) indices and ```Vector3```'s for the reference ground position and ground plane normal. It returns a ```Vector3``` for the projected position on the ground. The extended interface allows you to call the function with a Python sequence (list, tuple, ndarray, etc.) of row values, a Python sequence of column values (same length as row sequence), and ```Vector3```'s for a single reference ground position and normal vector. The return is a Python list of ```Vector3```'s for each projected ground position.

The use case is fast projection of many (row,column) pairs from Python while avoiding the performance overhead of Python looping.

Added a unit test that checks (1) functionality of the new interface and (2) that the projected return values for a list of inputs matches the projected return for each list entry individually.


modules/python/scene/source/scene.i
```diff
@@ -91,3 +91,51 @@
...
+    {
+      PyErr_SetString(PyExc_TypeError,"Expecting a sequence of row indices.");
+      return NULL;
+    }
+    if (!PySequence_Check(col_seq))
+    {
+      PyErr_SetString(PyExc_TypeError,"Expecting a sequence of column indices.");
+      return NULL;
+    }
+    int N = PyObject_Length(row_seq);
```
asylvest: Use `Py_ssize_t` here and below

dstarinshak: Changed ‘int’ to ‘Py_ssize_t’ throughout update.

PR #36, opened 2015-12-16
Remove formal param warnings

JonathanMeans:
This will take care of all the frivolous one, except from external places. Will try to deal with a common one from Nitro shortly.

### DIFF
--- a/six/modules/c++/six.sicd/source/ComplexXMLParser04x.cpp
+++ b/six/modules/c++/six.sicd/source/ComplexXMLParser04x.cpp
@@ -330,14 +330,14 @@ XMLElem ComplexXMLParser04x::convertHPBWToXML(
 }
 
 XMLElem ComplexXMLParser04x::convertAntennaParamArrayToXML(
-    const std::string& name,
+    const std::string& /*name*/,
     const GainAndPhasePolys* array,
     XMLElem apXML) const
 {
     //! optional field in 0.4
     if (array)
     {
-        XMLElem arrXML = newElement(name, apXML);
+        XMLElem arrXML = newElement("Array", apXML);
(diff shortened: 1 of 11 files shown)


six/modules/c++/six.sicd/source/ComplexXMLParser04x.cpp
```diff
@@ -337,7 +337,7 @@ XMLElem ComplexXMLParser04x::convertAntennaParamArrayToXML(
     //! optional field in 0.4
     if (array)
     {
-        XMLElem arrXML = newElement("Array", apXML);
+        XMLElem arrXML = newElement(name, apXML);
```
asylvest: This is not right... if you look up a level to `ComplexXMLParser::convertAntennaToXML()`, the name that is sent down is `Tx`, `Rcv`, or `TwoWay`, not `Array`.  It looks like the reason `name` is sent down is so that `ComplexXMLParser10x::convertAntennaParamArrayToXML()` can throw and include the name in there... `ComplexXMLParser04x` doesn't need to do that (that field must be optional), so instead you want to just remove the name of the variable in `ComplexXMLParser04x.cpp`

asylvest:
Everything looks good except for the one comment above.  Did you track down the NITRO warning?

JonathanMeans:
I changed the name thing.

On Nitro, the warnings that keep coming up are from this file: https://github.com/mdaus/nitro/blob/master/modules/c/nitf/source/BandSource.c

There are about four functions (MemorySource_contigRead, MemorySource_offsetRead, MemorySource_getSize, IOSource_getSize) that take pointers to nitf_Errors, then don't do anything with them. I was going to try to work out whether something should have been done, or if it's a convention thing, or whatever it is after I get my Manticore issue fixed.

asylvest:
Looks good

PR #38, opened 2016-01-11
Move modules/c++/six.sicd/tests/test_sicd_parsing.cpp to modules/c++/samples/test_six_parsing.cpp

JonathanMeans:


six/modules/c++/samples/test_six_parsing.cpp
```diff
@@ -111,14 +116,22 @@ void XMLVerifier::verify(const std::string& pathname) const
...
-                                                 six::DataType::COMPLEX,
+                                                 mType,
                                                  mSchemaPaths,
                                                  mLog));
 
     // Write it back out - this verifies both that the XML we write validates
     // against the schema and that our parser writes it without errors
-    six::sicd::ComplexXMLControl xmlControl;
-    std::auto_ptr<xml::lite::Document> xmlDoc(xmlControl.toXML(data.get(),
+    six:: XMLControl* xmlControl;
```
asylvest: Note that this will leak (who deletes `xmlControl`?).  That's why we try to use smart pointers everywhere... see if there's a way you could refactor `six::parseData()` such that you didn't need this check here  and/or if there's a way you can easily do this differently.  I'm trying to avoid the caller needing to specify SICD vs. SIDD on the command line (due to my incredible laziness).  Let me know if you want to catch up on options.

six/modules/c++/samples/test_six_parsing.cpp
```diff
@@ -18,7 +18,7 @@
  * License along with this program; If not,
  * see <http://www.gnu.org/licenses/>.
  *
- * Simple program to round-trip SICD XML and verify the resulting XML matches
+ * Simple program to round-trip SI*D XML and verify the resulting XML matches
```
asylvest: Change `SI*D` to `SICD or SIDD`

six/modules/c++/samples/test_six_parsing.cpp
```diff
@@ -143,16 +156,17 @@ int main(int argc, char** argv)
     try
     {
         // Parse the command line
-        if (argc < 2)
+        if (argc < 3 || !(strcmp(argv[1], "SICD") && !strcmp(argv[1], "SIDD")))
         {
             std::cerr << "Usage: " << sys::Path::basename(argv[0])
+                      << "\"SICD\" | \"SIDD\"\n"
                       << " <SICD XML pathname #1>"
```
asylvest: Change to `SICD/SIDD XML pathname`

asylvest:
There is probably some clever way to do this to avoid some of the code duplication, but it is all simple stuff duplicated so not the end of the world... looks good.

PR #40, opened 2016-01-19
exposing the functionality to project a slant plane pixel into the ground plane latLong

mbojrab:
Needed this same functionality on the MIOD integration, so I thought it could be a good candidate for exposing to all. I've moved the contents of the functions back into the cpp to make the header more readable.

### DIFF
--- a/six/modules/c++/six.sicd/include/six/sicd/CropUtils.h
+++ b/six/modules/c++/six.sicd/include/six/sicd/CropUtils.h
@@ -29,11 +29,59 @@
 #include <types/RowCol.h>
 #include <scene/Types.h>
 #include <six/NITFReadControl.h>
+#include <six/sicd/ComplexData.h>
 
 namespace six
 {
 namespace sicd
 {
+/*!
+ * Projects a slant plane pixel into ground plane LLA coordinates
+ */
+class PixelToLatLon
+{
+public:
+    PixelToLatLon(const six::sicd::ComplexData& data,
+                  const scene::SceneGeometry& geom,
+                  const scene::ProjectionModel& projection) :
+        mGeom(geom),
+        mProjection(projection),
+        mOffset(data.imageData->scpPixel.row -
+                        static_cast<double>(data.imageData->firstRow),
+                data.imageData->scpPixel.col -
+                        static_cast<double>(data.imageData->firstCol)),
+        mSampleSpacing(data.grid->row->sampleSpacing,
+                       data.grid->col->sampleSpacing),
+        mGroundPlaneNormal(mGeom.getReferencePosition())
+    {
+        mGroundPlaneNormal.normalize();
+    }
+
+    scene::LatLon operator()(size_t row, size_t col) const
+    {
+        const types::RowCol<double> imagePt(
+                (row - mOffset.row) * mSampleSpacing.row,
+                (col - mOffset.col) * mSampleSpacing.col);
+
+        double timeCOA(0.0);
+        const scene::Vector3 groundPt =
+                mProjection.imageToScene(imagePt,
+                                         mGeom.getReferencePosition(),
+                                         mGroundPlaneNormal,
+                                         &timeCOA);
+
+        const scene::LatLonAlt latLon(scene::Utilities::ecefToLatLon(groundPt));
+        return scene::LatLon(latLon.getLat(), latLon.getLon());
+    }
+
+private:
+    const scene::SceneGeometry& mGeom;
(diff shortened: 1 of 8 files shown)


six/modules/c++/six.sicd/include/six/sicd/CropUtils.h
```diff
@@ -29,11 +29,32 @@
...
 #include <scene/Types.h>
 #include <six/NITFReadControl.h>
+#include <six/sicd/ComplexData.h>
 
 namespace six
 {
 namespace sicd
 {
+/*!
+ * Projects a slant plane pixel into ground plane LLA coordinates
```
asylvest: It's just lat, lon (not lat, lon, alt, even though we have that internally initially), so change comment that says LLA to something like lat/lon

six/modules/c++/six.sicd/include/six/sicd/CropUtils.h
```diff
@@ -29,11 +29,32 @@
...
+ */
+class PixelToLatLon
```
asylvest: Put this in its own header rather than in CropUtils

six/modules/c++/six.sicd/source/CropUtils.cpp
```diff
@@ -166,6 +121,40 @@ namespace six
...
+    mSampleSpacing(data.grid->row->sampleSpacing,
+                   data.grid->col->sampleSpacing),
+    mGroundPlaneNormal(mGeom.getReferencePosition())
+{
+    mGroundPlaneNormal.normalize();
+}
+
+scene::LatLon PixelToLatLon::operator()(size_t row, size_t col) const
+{
+    const types::RowCol<double> imagePt(
```
asylvest: Take a look at `ComplexData::pixelToImagePoint()`.  Skimming over things, I'm pretty sure you could use that to get imagePt and remove `mOffset` and `mSampleSpacing` member variables (technically it's a little less efficient as you'd be re-computing some stuff but it'll be trivial compared to the `imageToScene()` call).

I added `pixelToImagePoint()` relatively recently... looks like I didn't notice that I could have updated this code to use it.

I do think `PixelToLatLon` should stay as its own class rather than be a member of `ComplexData` though so that `ComplexData` doesn't know about `ProjectionModel`.  I could be convinced otherwise but I think this standalone class is easier to use for this anyway.

six/modules/c++/six.sicd/include/six/sicd/CropUtils.h
```diff
@@ -29,11 +29,32 @@
...
+{
+public:
+    PixelToLatLon(const six::sicd::ComplexData& data,
+                  const scene::SceneGeometry& geom,
+                  const scene::ProjectionModel& projection);
+
+    scene::LatLon operator()(size_t row, size_t col) const;
+
+private:
+    const scene::SceneGeometry& mGeom;
```
asylvest: Note that `mGeom` and `mProjection` are just held by const ref.  This was definitely safe when in the unnamed namespace since we knew exactly how it was being used but now it can be screwed up.  I'd hold `mGeom` by value (copying it is trivially cheap) but you can't copy `mProjection` (it's an abstract base class).  For the time being I'd say still hold it by const ref but make it clear in the constructor comments (the alternative is to pass it in by shared pointer which is safer but will force some rippling API changes... I don't have a defensible argument against doing this though).

six/modules/c++/six.sicd/include/six/sicd/CropUtils.h
```diff
@@ -29,11 +29,32 @@
...
 {
 namespace sicd
 {
+/*!
+ * Projects a slant plane pixel into ground plane LLA coordinates
+ */
+class PixelToLatLon
+{
+public:
+    PixelToLatLon(const six::sicd::ComplexData& data,
```
asylvest: Add Doxygen comments to both constructor and `operator()`

asylvest:
OK last thing... since CropUtils wanted lat/lon, you wanted lat/lon/alt, and someone might want just ECEF that day, it seems like this `PixelToLLA` class should provide methods that do all three.  A base `toECEF()` implementation that does 95% of the work, then `toLatLon()` and `toLatLonAlt()` or `toLLA()` to get you the others for convenience.

Only thing is then it's not a functor anymore so needs to be renamed and has to be a noun.  `SlantPlanePixelTransformer`?

mbojrab:
Good point, I've made those updates and updated cropUtils to use the appropriate functionality.

asylvest:
Looks good

PR #44, opened 2016-01-26
Regression testing

asylvest:


asylvest:
@JonathanMeans can you look into and resolve the merge conflicts?

JonathanMeans:
Resolved.

six/modules/python/six.sicd/tests/test_create_sicd_xml.py
```diff
@@ -34,686 +35,777 @@
...
-root.prettyPrint(fos)
-
-# These should match #
-if filecmp.cmp(origPathname, newPathname):
-    print 'Round trip succeeded!'
-else:
-    print 'Round trip failed'
-
-# If we made it to here, the read side appears to be working properly too
+def initCollectionInfo(cmplx):
```
asylvest: Did line endings change or why does every line show up as a diff?  Seems like the old file already had Unix line endings...

six/modules/python/six.sicd/source/six_sicd.i
```diff
@@ -39,20 +39,83 @@ using namespace six;
...
     throw except::BadCastException();
   }
-  else 
+  else
   {
     return complexData;
   }
 }
 
+void writeNITF(std::string fileName, std::vector<std::string> schemaPaths,
```
asylvest: Pass in strings and vector of strings by const ref.  Also, use `pathname` rather than `fileName` throughout (we use `pathname` when it's the directory plus filename and `filename` when it doesn't contain the directory.

Pass `data` in by const ref rather than non-const pointer.

six/modules/python/six.sicd/source/six_sicd.i
```diff
@@ -39,20 +39,83 @@ using namespace six;
...
+        six::sicd::ComplexData* data, long long imageAdr);
+
+void writeNITF(std::string fileName, std::vector<std::string> schemaPaths,
+        six::sicd::ComplexData* data, long long imageAdr)
+{
+    std::complex<float>* image = reinterpret_cast< std::complex<float>* >(
```
asylvest: This can be a `const std::complex<float>*`

six/modules/python/six.sicd/source/six_sicd.i
```diff
@@ -39,20 +39,83 @@ using namespace six;
...
+            imageAdr);
+
+    six::XMLControlFactory::getInstance().addCreator(
+            six::DataType::COMPLEX,
+            new six::XMLControlCreatorT<six::sicd::ComplexXMLControl>());
+
+    six::Container container(six::DataType::COMPLEX);
+    std::auto_ptr<logging::Logger> logger(logging::setupLogger("out"));
+
+    container.addData(data);
```
asylvest: Take a look at the API but I'm pretty sure the container takes ownership of the data.  I'm assuming that's problematic up a level... you probably need to pass in `data.clone()`.

asylvest:
In the future, to resolve those merge conflicts, you should be able to do it in the branch rather than merging it into master... the goal is to always get to the point where a branch has no conflicts and can be one-click merged into master.

So, @JonathanMeans go ahead and make the changes in the comments above on a branch and open up a new pull request for it.

six/modules/python/six.sicd/tests/test_create_sicd_xml.py
JonathanMeans: I did a rather large refactor on that file. Originally, creating the complex data all happened from the main program. I split each section of the SICD into its own function so it would be easier to work with.

PR #45, opened 2016-01-28
Fix bug in parsing RGB LUT

asylvest:


asylvest:
Looks good @JonathanMeans... you could add `const` here and there to the couple intermediate values, but that's just a general style thing to keep in mind so this is good as-is.

PR #48, opened 2016-02-12
Fix object slicing issue with Annotations, clean up some warnings

JonathanMeans:
Had to leave in a couple empty default ctors, since those classes now have custom copy ctors.

### DIFF
--- a/six/modules/c++/six.sidd/include/six/sidd/SFA.h
+++ b/six/modules/c++/six.sidd/include/six/sidd/SFA.h
@@ -282,7 +282,7 @@ public:
         return new SFAMultiLineString(*this);
     }
 
-    std::vector<mem::ScopedCopyablePtr<SFALineString> > elements;
+    std::vector<mem::ScopedCloneablePtr<SFALineString> > elements;
     static const char TYPE_NAME[];
 };
 
--- a/six/modules/c++/six.sidd/include/six/sidd/SFA.h
+++ b/six/modules/c++/six.sidd/include/six/sidd/SFA.h
@@ -43,7 +43,7 @@ public:
 
 protected:
     std::string mType;
-    SFATyped(std::string typeName) :
+    SFATyped(const std::string& typeName) :
         mType(typeName)
     {
     }
@@ -51,13 +51,13 @@ protected:
 
 struct SFAGeometry : public SFATyped
 {
-public:
-    virtual ~SFAGeometry()
+    virtual SFAGeometry* clone() const
     {
+        return new SFAGeometry(*this);
     }
 
 protected:
-    SFAGeometry(std::string typeName) :
+    SFAGeometry(const std::string& typeName) :
         SFATyped(typeName)
     {
     }
@@ -88,8 +88,9 @@ struct SFAPoint : public SFAGeometry
     {
     }
 
-    virtual ~SFAPoint()
+    virtual SFAPoint* clone() const
     {
+        return new SFAPoint(*this);
     }
 
     static const char TYPE_NAME[];
@@ -98,12 +99,8 @@ struct SFAPoint : public SFAGeometry
 //! Abstract type
 struct SFACurve : public SFAGeometry
 {
-public:
-    virtual ~SFACurve()
-    {
-    }
 protected:
-    SFACurve(std::string typeName) :
+    SFACurve(const std::string& typeName) :
         SFAGeometry(typeName)
     {
     }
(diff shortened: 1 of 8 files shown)


asylvest:
Yep the clone is what's needed like you have it.  We also have another type in mem called `ScopedCloneablePtr`.  Take a look at it; I think you can switch to that and get rid of the copy constructors.

JonathanMeans:
That makes things easier. Just pushed up the change.

six/modules/c++/six.sidd/include/six/sidd/SFA.h
```diff
@@ -88,8 +88,9 @@ struct SFAPoint : public SFAGeometry
     {
```
asylvest: As long as we're touching this code, update these constructors to initialize these members via initializer list instead of in the constructor body

six/modules/c++/six.sidd/include/six/sidd/SFA.h
```diff
@@ -43,7 +43,7 @@ struct SFATyped
 
 protected:
     std::string mType;
-    SFATyped(std::string typeName) :
+    SFATyped(const std::string& typeName) :
```
asylvest: Add a pure virtual `clone()` method in the base class

six/modules/c++/six.sidd/include/six/sidd/SFA.h
```diff
@@ -205,8 +201,10 @@ struct SFAPolyhedralSurface : public SFASurface
...
     {
     }
-    virtual ~SFAPolyhedralSurface()
+
+    virtual SFAPolyhedralSurface* clone() const
     {
+        return new SFAPolyhedralSurface(*this);
     }
 
     std::vector<mem::ScopedCopyablePtr<SFAPolygon> > patches;
```
asylvest: This (and SFAMultiPolygon) need to still be switched to be a vector of scoped cloneables right?  The vector of scoped copyables of SFAPoint is ok as-is

asylvest: Actually looks like there are several other spots... you'll want to look at each one to see if it's going to slice or not.

JonathanMeans: Made the edits. I’m pretty sure the ScopedCopyablePtr<SFALineString> on line 285 can stay as is, because there are no ‘extra’ fields in the two subclasses to be sliced off, just a different TYPE_NAME value.

asylvest: Hmm, so I think you're right.  I'd have to go dig through some C++ documentation to say for sure.  But just for code clarity, let's say any time something is a base class where things are inheriting from it, let's use a vector of cloneables.  I think line 285 is the only remaining one that needs to get changed.

JonathanMeans: Alright, it’s changed.

six/modules/c++/six.sidd/include/six/sidd/SFA.h
```diff
@@ -68,16 +70,18 @@ struct SFAPoint : public SFAGeometry
...
 
     SFAPoint() :
-        SFAGeometry(TYPE_NAME)
+        SFAGeometry(TYPE_NAME), x(0), y(0),
+        z(::six::Init::undefined<double>()),
+        m(::six::Init::undefined<double>())
     {
-        x = 0;
-        y = 0;
         z = ::six::Init::undefined<double>();
```
asylvest: z and m can be removed from the constructor body.  Also, put x and y on their own lines

six/modules/c++/six.sidd/include/six/sidd/SFA.h
```diff
@@ -68,16 +70,18 @@ struct SFAPoint : public SFAGeometry
...
         m = ::six::Init::undefined<double>();
     }
     
     SFAPoint(double _x, double _y) :
-        SFAGeometry(TYPE_NAME), x(_x), y(_y)
+        SFAGeometry(TYPE_NAME), x(_x), y(_y),
+        z(::six::Init::undefined<double>()),
+        m(::six::Init::undefined<double>())
     {
         z = ::six::Init::undefined<double>();
```
asylvest: Ditto

asylvest:
And one last thing - remove the extra newline at the end of the file

six/modules/c++/six.sidd/include/six/sidd/SFA.h
JonathanMeans: Do you mean one line for both x and y, or x on one line and y on another?

asylvest:
x and y each on their own line (we don't do this everywhere, but for new/updated code we should)

JonathanMeans:
Done

asylvest:
Good deal

PR #57, opened 2016-04-04
Regression testing

asylvest:
@JonathanMeans


six/modules/c++/samples/test_create_sidd_from_mem.cpp
```diff
@@ -1773,17 +1774,498 @@ void initProcessorInformation(
...
+    productCreation.classification.declassDate.reset(new six::DateTime());
+    productCreation.classification.declassEvent = "N/A";
+    productCreation.classification.declassException = "25X1";
+
+    //SIDD 1.0 only
+    productCreation.classification.exemptedSourceType = "X8";
+    productCreation.classification.exemptedSourceDate.reset(new six::DateTime());
+
+    /*
+    This is SIDD_1.1 stuff.
```
asylvest: Keep all the SIDD 1.1 stuff in a branch for now... you might want to make a note of what all such changes are and then once this is merged into master, merge master into the SIDD 1.1 branch and put this stuff in.

six/modules/c++/samples/test_create_sidd_from_mem.cpp
```diff
@@ -1773,17 +1774,498 @@ void initProcessorInformation(
...
+    display.monitorCompensationApplied.reset(new six::sidd::MonitorCompensationApplied());
+    display.monitorCompensationApplied->gamma = 5.9;
+    display.monitorCompensationApplied->xMin = 0.87;
+
+}
+void initDisplay(six::sidd::Display& display, const std::string& lutType,
+        const std::string& version)
 {
-    if (argc != 2 && argc != 3)
+    if (version == "1.0.0")
```
asylvest: Ditto here - needs to stay in SIDD 1.1 branch

six/modules/c++/samples/test_create_sidd_from_mem.cpp
```diff
@@ -1773,17 +1774,498 @@ void initProcessorInformation(
...
+    system->primeMeridian.longitude = 11.11;
+    system->linearUnit = "Unit";
+
+    annotation->objects.resize(1);
+    annotation->objects[0].reset(new six::sidd::SFAPoint(2.4, 3.5, 4.55, 5.757));
+    annotations.push_back(annotation);
+}
+
+void populateData(six::sidd::DerivedData& siddData, const std::string&
+        lutType, const bool smallImage, const std::string& version)
```
asylvest: Input parameters passed by value don't need to be `const` (since the caller knows you can't modify them outside the scope of this function), so just send `smallImage` in as a `bool`

six/modules/c++/samples/test_create_sidd_from_mem.cpp
```diff
@@ -1773,17 +1774,498 @@ void initProcessorInformation(
...
+    argParser.addArgument("xml", "Optional SICD .xml file", cli::STORE,
+            "sicdXML", "sicd-xml", 0, 1);
+
+    //We'll get this from command line args once it matters
+    const std::string version("1.0.0");
 
     try
     {
+        std::auto_ptr<cli::Results> options(argParser.parse(argc,
+                (const char**) argv));
```
asylvest: Don't need the `const char**` cast here (old code might have been doing this b/c we used to not have an overloading)

six/modules/c++/samples/test_create_sidd_from_mem.cpp
```diff
@@ -1835,147 +2334,58 @@ int main(int argc, char** argv)
...
-        parent->information->collectionDateTime = six::DateTime();
-        parent->information->radarMode = RadarModeType::SPOTLIGHT;
-        parent->information->sensorName = "";
-        siddData->exploitationFeatures->product.resolution.row = 0;
-        siddData->exploitationFeatures->product.resolution.col = 0;
+            populateData(*siddData, lutType, smallImage, version);
+            container.addData(siddData->clone());
+            if (!smallImage)
+            {
+                buffers[ii] = (UByte*) IMAGE.data;
```
asylvest: Use C++ rather than C cast, so here `reinterpret_cast`

six/modules/c++/samples/test_create_sidd_from_mem.cpp
```diff
@@ -1835,147 +2334,58 @@ int main(int argc, char** argv)
...
+            }
+            else
+            {
+                buffers[ii] = (UByte*) smallData;
```
asylvest: Use C++ rather than C cast, so here `reinterpret_cast`

six/modules/c++/six.sicd/source/ComplexXMLParser.cpp
```diff
@@ -1316,6 +1316,22 @@ void ComplexXMLParser::parseImageFormationFromXML(
...
+    XMLElem areaElem = getOptional(getFirstAndOnly(imageFormationXML->getParent(), "RadarCollection"), "Area");
+    if (areaElem) 
+    {
+        XMLElem planeElem = getOptional(areaElem, "Plane");
+        if (planeElem) 
+        {
+            segmentElem = getOptional(planeElem, "SegmentList");
+        }
+    }
+    if (!tmpElem && segmentElem)
```
asylvest: Good catch.  Rather than re-parsing the XML though, you can just check if sicd->radarCollection->area->plane->segmentList exists (with the NULL checks along the way obviously).  I would say pass in a const ref to a `RadarCollection` into this function.  That way it is hopefully obvious up a level that we can't arbitrarily re-order calling this vs. `parseRadarCollectionFromXML()`.

six/modules/c++/six.sicd/source/ComplexXMLParser.cpp
```diff
@@ -1316,6 +1316,22 @@ void ComplexXMLParser::parseImageFormationFromXML(
...
+    {
+        std::cerr <<
+            " !!! " << std::endl;
+        throw except::Exception(Ctxt("SegmentList must be included when a radarCollection->area->plane::segmentList is included."));
```
asylvest: Remove `cerr`.  Also, exception is referring to SegmentList when it should be referring to SegmentIdentifier.  Reword to something like "SegmentIdentifier must be included when RadarCollection.Area.Plane.SegmentList is present".

six/modules/c++/six.sicd/source/ComplexXMLParser04x.cpp
```diff
@@ -160,11 +160,23 @@ XMLElem ComplexXMLParser04x::convertImageFormationToXML(
     //! this segment was recreated completely because the ordering of
     //! a lot of the variables has been updated
     XMLElem imageFormationXML = newElement("ImageFormation", parent);
-
+    XMLElem segmentElem = NULL;
```
asylvest: Similarly here, use the C++ structure to check for existence and fix exception message

six/modules/c++/six.sicd/source/ComplexXMLParser10x.cpp
```diff
@@ -181,6 +181,21 @@ XMLElem ComplexXMLParser10x::convertImageFormationToXML(
     createDouble("MinProc", imageFormation->txFrequencyProcMin, txFreqXML);
     createDouble("MaxProc", imageFormation->txFrequencyProcMax, txFreqXML);
 
+    XMLElem segmentElem = NULL;
```
asylvest: Similarly here, use the C++ structure to check for existence and fix exception message

six/modules/c++/six.sidd/include/six/sidd/Annotations.h
```diff
@@ -40,6 +40,17 @@ struct Annotation
         identifier(""), spatialReferenceSystem(NULL)
     {
     }
+
+    Annotation(const Annotation& other)
```
asylvest: Merge master into this branch - this isn't needed now

six/modules/c++/six.sidd/include/six/sidd/SFA.h
```diff
@@ -56,6 +56,11 @@ struct SFAGeometry : public SFATyped
     {
     }
 
+    virtual SFAGeometry* clone() const
```
asylvest: These are all in master now too right?

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+void generateData(std::complex<float>* data)
+{
+    for (size_t ii = 0; ii < DATA_LENGTH; ++ii)
+    {
+        data[ii] = std::complex<float>((float)ii, (float)ii);
+    }
+}
+
+six::sidd::DerivedData* createData()
```
asylvest: Use an `auto_ptr` rather than raw pointer

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -0,0 +1,195 @@
...
+    planeProjection->productPlane.rowUnitVector = six::Vector3(0.0);
+    planeProjection->productPlane.colUnitVector = six::Vector3(0.0);
+    derivedData->measurement->arpPoly = six::PolyXYZ(0);
+    derivedData->measurement->arpPoly[0] = six::Vector3(0.0);
+    derivedData->display.reset(new six::sidd::Display());
+    derivedData->display->pixelType = six::PixelType::MONO16I;
+    derivedData->setNumRows(10);
+    derivedData->setNumCols(40);
+    derivedData->geographicAndTarget.reset(new six::sidd::GeographicAndTarget(six::RegionType::GEOGRAPHIC_INFO));
+    //derivedData->geographicAndTarget->imageCorners.reset(new six::LatLonCorners());
```
asylvest: If these two lines aren't needed, just remove them rather than keep them commented out

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -0,0 +1,195 @@
...
+    six::sidd::Collection* parent =
+        derivedData->exploitationFeatures->collections[0].get();
+    parent->information->resolution.rg = 0;
+    parent->information->resolution.az = 0;
+    parent->information->collectionDuration = 0;
+    parent->information->collectionDateTime = six::DateTime();
+    parent->information->radarMode = six::RadarModeType::SPOTLIGHT;
+    parent->information->sensorName.clear();
+    parent->geometry.reset(new six::sidd::Geometry());
+    //parent->geometry->dopplerConeAngle = 45.8;
```
asylvest: Ditto - remove if unneeded

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -0,0 +1,195 @@
...
+    return derivedData;
+}
+
+void write(six::UByte* data, bool useStream, bool byteSwap)
+{
+    six::Container container(six::DataType::DERIVED);
+    container.addData(createData());
+
+    six::NITFWriteControl writer;
+    writer.getOptions().setParameter(six::WriteControl::OPT_BYTE_SWAP, (int)byteSwap);
```
asylvest: Use `static_cast` rather than C-style cast

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -0,0 +1,195 @@
...
+    writer.initialize(&container);
+
+    if (useStream)
+    {
+        io::ByteStream stream;
+        stream.write((sys::byte*)data, DATA_SIZE_IN_BYTES);
```
asylvest: Use `reinterpret_cast`.  We really need to change the io module API to send in a `void*` instead to avoid these annoying casts but since the method is virtual we have to be a little careful that we don't break existing classes that inherit from it, so we won't bite that off yet.

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    const size_t numRows = inData->getNumRows();
+    const size_t numCols = inData->getNumCols();
+
+    six::Region region;
+    region.setStartRow(0);
+    region.setStartCol(0);
+    region.setNumRows(numRows);
+    region.setNumCols(numCols);
+    six::UByte buffer[DATA_SIZE_IN_BYTES];
```
asylvest: Why is this extra buffer needed?  Can't you read straight into `data`?

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -0,0 +1,195 @@
...
+bool run(bool useStream = false, bool byteswap = false)
+{
+    std::complex<float> complexData[DATA_LENGTH];
+    generateData(complexData);
+
+    six::UByte* sourceData = reinterpret_cast<six::UByte*>(&complexData[0]);
+    six::UByte testData[DATA_SIZE_IN_BYTES];
+    memcpy(testData, sourceData, DATA_SIZE_IN_BYTES);
+
+    if (!sys::isBigEndianSystem() && !byteswap || sys::isBigEndianSystem() && byteswap)
```
asylvest: This will do what you want because `&&` will win over `||` but for code clarity add some parens:
`if ((!sys::isBigEndianSystem() && !byteswap) || (sys::isBigEndianSystem() && byteswap))`

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -0,0 +1,195 @@
...
+        six::DataType::DERIVED,
+        new six::XMLControlCreatorT<
+        six::sidd::DerivedXMLControl>());
+
+    try
+    {
+        bool success = run(false, false) && run(true, false) &&
+            run(false, true) && run(false, false);
+
+        if (success)
```
asylvest: Shorten this to just `return (success ? 0 : 1))` and get rid of the `return 1` at the end

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -0,0 +1,195 @@
...
+    std::complex<float> complexData[DATA_LENGTH];
+    generateData(complexData);
+
+    six::UByte* sourceData = reinterpret_cast<six::UByte*>(&complexData[0]);
+    six::UByte testData[DATA_SIZE_IN_BYTES];
+    memcpy(testData, sourceData, DATA_SIZE_IN_BYTES);
+
+    if (!sys::isBigEndianSystem() && !byteswap || sys::isBigEndianSystem() && byteswap)
+    {
+        sys::byteSwap(sourceData, BYTES_PER_PIXEL, DATA_SIZE_IN_BYTES / BYTES_PER_PIXEL);
```
asylvest: Eh, there is a mismatch in here.  You are writing out a SIDD of type `six::PixelType::MONO16I` which means you are writing out 16-bit integer data.  You should be using arrays of `sys::Int16_T` rather than `std::complex<float>`.  This is also why `BYTES_PER_PIXEL` is 2... change that to `sizeof(sys::Int16_T)` throughout.

six/modules/c++/six/tests/test_polarizationType_conversions.cpp
```diff
@@ -0,0 +1,49 @@
+#include <six/Enums.h>
```
asylvest: Add banner up top

six/modules/c++/six/tests/test_polarizationType_conversions.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <six/Utilities.h>
```
asylvest: Actually make this a unit test (under six/unittests) and use `TestCase.h` to do these tests

six/modules/python/six.sicd/tests/test_create_sicd_xml.py
```diff
@@ -677,36 +717,45 @@ def initRMA():
...
-def initINCA():
+def initINCA(rma):
     inca = makeScopedCopyableINCA()
     inca.timeCAPoly = Poly1D(3)
     for i in range(3):
 	inca.timeCAPoly[i] = i
     inca.rangeCA = 5
     inca.freqZero = 10
-    inca.dopplerRateScaleFactorPoly = Poly2D(3, 3)
+    inca.dopplerRateScaleFactorPoly = Poly2D(0, 3)
```
asylvest: Why was this changed..?

six/modules/python/six.sicd/tests/test_create_sicd_xml.py
```diff
@@ -718,36 +767,50 @@ def initData(cmplx, includeNITF):
...
-    initRMA()
-    initRMCR()
-    initINCA()
-    initRgAzComp()
+    if alg == 'PFA':
+        cmplx = initPFA(cmplx)
+    elif alg == 'RMA':
+        cmplx = initRMA(cmplx, version, imageType)
+    elif alg == 'RGAZCOMP':
+        if not version.startswith('1'): #TODO: verify version cutoff
```
asylvest: Can you verify this todo?

six/modules/python/six/tests/makeSICDRegressionFiles.py
```diff
@@ -0,0 +1,74 @@
...
+
+    home = utils.findSixHome()
+    if not os.path.isdir(os.path.join(home, 'regression_files', 'six.sicd')):
+        os.makedirs(os.path.join(home, 'regression_files', 'six.sicd'))
+    for args in product(sicdVersions, formationAlgs, imageTypes):
+        if args[1] != 'RMA':
+            args = (args[0], args[1], '')
+        createNITFs(args[0], args[1], args[2], home)
+ 
+
```
asylvest: Extra newline

six/modules/python/six/tests/utils.py
```diff
@@ -0,0 +1,83 @@
...
+            if 'tests' in subdirs and 'bin' in subdirs:
+                return child
+
+def setPaths():
+    nitfPluginPath = os.path.join(installPath(), 'share', 'nitf',
+                                  'plugins')
+    sixSchemaPath = os.path.join(installPath(), 'conf', 'schema',
+                                 'six')
+
+    pythonPath = os.path.join(installPath(), 'lib', 'python2.7',
```
asylvest: Do a search for a `python*` directory here in case someone is using Python 2.6 (or for when we go to 3.x someday)

six/modules/python/six/tests/runTests.py
```diff
@@ -0,0 +1,56 @@
...
+
+print "Performing byte swap test"
+if subprocess.call([utils.executableName(os.path.join(
+        utils.installPath(), 'tests', 'six.sidd', 'test_byte_swap'))]) != 0:
+    print "Failed ByteSwap test in six.sidd/tests/test_byte_swap"
+    sys.exit(1)
+print "Byte swap test succeeded"
+
+print "All passed"
+sys.exit(0)
```
asylvest: Add a newline at end of file

six/modules/python/six/tests/runTests.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
```
asylvest: So is this my one-liner Python script I can run which will test everything?

asylvest:
I recommend merging master into this branch before starting to address any of these comments (this'll make a few of the mods on the branch no longer needed and will also resolve merge conflicts that are currently present).  It looks like maybe you've been trying to keep it up to date with the latest additions in master but somehow in a way that git got confused by... were you using git merge to do it (I would have thought/hoped that would have worked).

Anyway, let me know when it's good to look at again...

JonathanMeans:
I’m fuzzy on the details, but I recall doing something odd to sync up the branches for some reason. But merging in master this time worked time.

six/modules/python/six.sicd/tests/test_create_sicd_xml.py
JonathanMeans: Because version 0.4 of the schema expects it to look like a Poly1D. Although the doc says 2D, so now I’m thinking it might be a typo in the schema?

six/modules/python/six/tests/runTests.py
JonathanMeans: Yes.

I’ve made the changes, and also added all the unittests to the testing script. A few of them are crashing. I’m assuming I should go ahead and fix that?

JonathanMeans: I’ve fixed the tests, to a certain extent. Test_sicd_schemata fails instead of segfaults,  but it has a number of problems that will become much more sane to work with once the equality operators get merged in.

asylvest:
You had an earlier comment about how some seg faults would be easier to deal with once the equality operator pull request was merged in... when you get a chance can you resolve those and fix the merge conflicts?  Let me know when it's good to review again.

JonathanMeans:
So I’m having an issue with one of the tests, where if I round trip a SICD, 0.01 is getting added to each element of a certain Vector3. (10 vs. 10.01). The difference is surprisingly large, so I want to take a close look at what’s going to see if there’s something wrong somehow. But otherwise I may have to go back and make equality “approximately equal” if it’s going to be useful.

Then again, I’d imagine there would be situations where values need to be less than 0.01 apart….

JonathanMeans:
Disregard.

JonathanMeans:
It’s good.

asylvest:
All little stuff:

- ComplexXMLParser::parseImageFormationFromXML(): Generally we try to keep output parameters last, so move `obj` as the last parameter
- ComplexXMLParser10x.cpp line 190, ComplexXMLParser04x.cpp line 170, ComplexXMLParser.cpp line 1325: Keep lines to 80 characters... you can continue the string on multiple lines in a couple different ways.  Also, the exception message should refer to the fields in a similar way (uses XML elements separated with periods for `ImageFormation.SegmentIdentifier` but then C++ notation for radarCollection).  I'd say switch radarCollection to the XML element period notation.
- test_byte_swap.cpp line 34: Set to `sizeof(sys::Int16_T)` rather than `2`
- test_byte_swap.cpp line 48: Set classification to `U`
- test_byte_swap.cpp lines 173, 177, 181: Need a `return 1` for each of these
- test_byte_swap.cpp line 109: Extra newline
- test_byte_swap.cpp line 124: `region` never has a buffer assigned to it... Under the hood, `NITFReadControl` will allocate a buffer for you, which will then leak because no one deletes it (`Region`'s destructor doesn't do anything).  But, instead you can just call `region.setBuffer()` with `data`... then there's not an allocation, leak, or `memcpy()`
- test_byte_swap.cpp line 133: I don't think there's any need to have `sourceData`... why not just eliminate it, have `testData` be of type `sys::Int16_T`, and then either change the for loop to go over `DATA_LENGTH` or you could use `memcmp()` instead of the for loop.  I think avoiding the extra pointer will make the code a little more legible.
- test_byte_swap.cpp line 139: Rather than `DATA_SIZE_IN_BYTES / BYTES_PER_PIXEL`, just use `DATA_LENGTH`
- test_polarizationType_conversions.cpp: Replace camelcase with underscores for filenames, so rename to test_polarization_type_conversions.cpp
- test_create_sicd_xml.py line 740: Gotcha... yeah, so that was a typo in the SICD 0.4.0 schema that they fixed in 0.4.1, but as you probably ran into to be consistent the C++ will throw if you try to write out a 2D poly for 0.4.0.  So yeah changing this line makes sense, but then go ahead and move it into its own for loop rather than still inside the double nested for loop on line 744.  Also indent lines 743-745 another 4 spaces.
- test_create_sicd_xml.py lines 99-100: These reappeared as part of the merge I guess (they're not in master)

JonathanMeans:
I’ve made the changes, and updated to work with Python3 as well.

asylvest:
Just a couple more nits in `test_byte_swap.cpp`:
- Move `addCreator()` into try block
- Generally only send smart pointers into a function/class if they need to know they're using a smart pointer (like sending a `mem::SharedPtr` into a constructor because you're making sure the object lives long enough).  So in the case of `read()` and `write()`, they just need to know they're working with raw pointers - they don't care it happens to be due to a caller making a `mem::ScopedArray` above them - so just send in a raw pointer.  At that point, the raw pointer to `write()` can be const.
- No `return` needed at the end of `read()`

JonathanMeans:
Fixed.

asylvest:
Removed a couple extra `const` keywords... when passing an input pointer in, or casting a pointer, and you want it to be const, doing `const T* ptr` makes sense, but no need to do `const T* const ptr`... the latter is making the address `ptr` points to itself const, which means someone in the function can't do `ptr = NULL`.  Style-wise, that's equivalent to passing in an `int` as a `const int` just to make sure inside the function they don't reassign the value.  While that does technically do something, those kinds of `const`s over-complicate the code (very slightly) so we don't do them since the caller knows that even if the function does modify a value internally, it won't modify the caller's value.

PR #58, opened 2016-04-04
Add UNKNOWN option to PolarizationType and DualPolarizationType

JonathanMeans:
The FrameType was in enums.txt, but not in Enums.h. The FrameType used in the six code gets typedefed over from scene, so I just removed it from the .txt.

### DIFF
--- a/six/modules/c++/six.sidd/source/Utilities.cpp
+++ b/six/modules/c++/six.sidd/source/Utilities.cpp
@@ -419,7 +419,7 @@ six::PolarizationType _convertDualPolarization(six::DualPolarizationType pol,
     case six::DualPolarizationType::LHC_LHC:
         return six::PolarizationType::LHC;
     case six::DualPolarizationType::UNKNOWN:
-        return six::PolarizationType::UNKNOWN;
+        throw except::Exception(Ctxt("DualPolarizationType::UNKNOWN has no corresponding PolarizationType"));
     default:
         return six::PolarizationType::NOT_SET;
     }
--- a/six/modules/c++/six.sidd/source/Utilities.cpp
+++ b/six/modules/c++/six.sidd/source/Utilities.cpp
@@ -418,6 +418,8 @@ six::PolarizationType _convertDualPolarization(six::DualPolarizationType pol,
                         : six::PolarizationType::RHC;
     case six::DualPolarizationType::LHC_LHC:
         return six::PolarizationType::LHC;
+    case six::DualPolarizationType::UNKNOWN:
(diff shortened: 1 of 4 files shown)


asylvest:
@JonathanMeans I just looked through the code and schemas.  Please double-check this but I believe that...

C++ PolarizationSequenceType == SICD schema Polarization1Type and should include UNKNOWN and SEQUENCE.  RadarCollection.txPolarization uses this.  So this one was right previously.

C++ PolarizationType == SICD schema Polarization2Type == SIDD schema sicommon:PolarizationType and does not include UNKNOWN and SEQUENCE.  SICD's RadarCollection.txSequence.txStep.txPolarization and SIDD's TxRcvPolarization.txPolarization and .rcvPolarization use this.  So this one was right previously too (and so the new enum value should be removed).

C++ DualPolarizationType == SICD schema DualPolarizationType and should include UNKNOWN but did not previously, so this addition was correct.

Can you verify and update the code accordingly?  Also, since you have to update enums.txt with this and it looks like things were out of sync with it and Enums.h based on the other stuff you needed to add in... remove CornerIndex from enums.txt too as it is not needed and was not in Enums.h previously.

six/modules/c++/six.sidd/source/Utilities.cpp
```diff
@@ -418,6 +418,8 @@ six::PolarizationType _convertDualPolarization(six::DualPolarizationType pol,
                         : six::PolarizationType::RHC;
     case six::DualPolarizationType::LHC_LHC:
         return six::PolarizationType::LHC;
+    case six::DualPolarizationType::UNKNOWN:
```
asylvest: I think line endings are goofed up on these two lines (don't just do a dos2unix under this pull request as lots of lines in this file unfortunately are wrong)

JonathanMeans:
Ah, was looking at the 1.1 spec.

Since DualPolarization has UNKNOWN and Polarization doesn’t, in sidd/Utilites.cpp _convertDualPolarization, would DualPolarizationType::UNKNOWN map to PolarizationType::OTHER?

asylvest:
Hmm, good question.  To be safe, let's have it throw in that case... if we ever run into it, we can re-evaluate.

JonathanMeans:
Changes are made.

(I pulled up Utilities.cpp in Notepad++. It showed consistent CRLF line-endings in the area around the comment. Not sure what else I should be doing about that.)

asylvest:
Can you remove CornerIndex from enums.txt?

PR #69, opened 2016-05-05
include sio.lite-python in installation

JonathanMeans:


asylvest:
Added a note explaining why this was done since it's not technically a dependency of `six-python` but I agree it makes sense to do this so the test scripts work and it just takes a sec to build

PR #76, opened 2016-05-24
Add method to allow adding extra DES to NITFWriteControl

JonathanMeans:
This also hooks things up to the Python scripts that test things. Some tabs got converted to spaces as well.


six/modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -251,6 +251,13 @@ class Utilities
     * \return the unit vector normal to the ground plane
     */
     static Vector3 getGroundPlaneNormal(const ComplexData& data);
+
+    /*
+     * Read a Data object from a SICD XML file.
+     * \param xmlPath The path to the input XML file
+     * \return the Data object represented by the XML
+     */
+    static std::auto_ptr<Data> readXML(const std::string& xmlPath);
```
asylvest: If you're going to put this there for convenience, might as well make it return a `ComplexData` since you know that's what it'll be

asylvest: Rename `xmlPath` to `xmlPathname`.  We just pathname for the full path, including a filename.

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,252 @@
+/* =========================================================================
+* This file is part of six.sidd-c++
```
asylvest: banner should be six.sicd, not six.sidd

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,252 @@
...
+* see <http://www.gnu.org/licenses/>.
+*
+*/
+
+#include <six/NITFWriteControl.h>
+#include <six/Types.h>
+#include <six/XMLControlFactory.h>
+#include <six/sicd/Utilities.h>
+
+
```
asylvest: Put all these in an unnamed namespace

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+    if (!str::endsWith(argv[1], ".xml"))
+    {
+        std::string message = "Extension suggests input may not be an XML file. "
+            "Expecting '*.xml'.";
+        throw except::Exception(Ctxt(message));
+    }
+}
+
+sys::Int16_T* generateBandData(const six::Data& data)
```
asylvest: Use a vector rather than a raw pointer that gets captured in a smart pointer.  That way the caller can't accidentally leak memory (granted this is a test case so you know all the contexts it's used in)

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,252 @@
...
+{
+    size_t length = data.getNumRows() * data.getNumCols();
+    six::UByte* bandData =
+        new six::UByte[length * data.getNumBytesPerPixel()];
+
+    for (size_t ii = 0; ii < length; ++ii)
```
asylvest: Why not just treat it as a `sys::Int16_T` all the way through?  You can cast `ii` to that instead...

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,252 @@
...
+    des.getSubheader().getVersion().set("01");
+    des.getSubheader().getSecurityClass().set("U");
+
+    mem::SharedPtr<nitf::SegmentWriter> segmentWriter(new nitf::SegmentWriter);
+    writer.addAdditionalDES(segmentWriter);
+
+    std::string name(std::tmpnam(NULL));
+    try
+    {
+        writer.save(reinterpret_cast<const six::UByte*>(bandData.get()), name);
```
asylvest: Is this fixed now that this comment no longer applies?

six/modules/c++/six.sidd/include/six/sidd/Utilities.h
```diff
@@ -70,6 +70,13 @@ class Utilities
     static std::auto_ptr<scene::ProjectionModel>
     getProjectionModel(const DerivedData* data);
 
+    /*
+    * Read a Data object from a SIDD XML file.
+    * \param xmlPath The path to the input XML file
+    * \return the Data object represented by the XML
+    */
+    static std::auto_ptr<Data> readXML(const std::string& xmlPath);
```
asylvest: Same deal - return DerivedData

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,252 @@
...
+        return false;
+    }
+}
+
+int main(int argc, char** argv)
+{
+    try
+    {
+        validateArguments(argc, argv);
+        std::string xmlName(argv[1]);
```
asylvest: Can be `const`

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+        bool allPassed = true;
+        allPassed = addingNullSegmentWriterShouldThrow(xmlName) && allPassed;
```
asylvest: Can you easily add a test where you use the 5 and 283 length versions of [XML_DATA_CONTENT](https://github.com/mdaus/nitro/blob/master/modules/c/nitf/shared/XML_DATA_CONTENT.c)?  I'm not actually sure how you tell NITRO to do this, so if it's not straightforward, you can punt for now.

asylvest: Try referring to the DES as XML_DATA_CONTENT_005 and see if that works...

six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -937,6 +937,14 @@ void NITFWriteControl::addDataAndWrite(
                                             false);
         deWriter.attachSource(segSource);
     }
+
+    int deWriterIndex = mContainer->getNumData();
+    for (size_t ii = 0; ii < mSegmentWriters.size(); ++ii)
+    {
+        mWriter.setDEWriteHandler(deWriterIndex, mSegmentWriters[ii]);
```
asylvest: Can save a line with `mWriter.setDEWriteHandler(deWriterIndex++, mSegmentWriters[ii]);`

six/modules/python/six/tests/runTests.py
```diff
@@ -47,6 +48,10 @@
     if checkNITFs.run() == False:
         print("test in checkNITFS.py failed")
         sys.exit(1)
+
+    if runMiscTests.run() == False:
+    # Tests should report their own errors
+        sys.exit(1)
 else:
     print('Warning: skipping the bulk of the test suite, as Python modules ' +
           'are by default disabled on Solairs')
```
asylvest: Just noticed typo - should be 'Solaris'

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
JonathanMeans: What does this do?

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
JonathanMeans: My thought process is that if the test breaks later, that means there’s a null pointer that’s not triggering an exception, so that branch is unlikely to happen even during a failure. But I guess it is a rather useless comment.

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,252 @@
...
+    std::string name(std::tmpnam(NULL));
+    try
+    {
+        writer.save(reinterpret_cast<const six::UByte*>(bandData.get()), name);
+        // Actually, the "default" behavior is to segfault...
+        std::cerr << "Test failed" << std::endl;
+        std::remove(name.c_str());
+        return false;
+    }
+    catch (except::Exception&)
```
asylvest: Can add a `const` here

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
JonathanMeans: Never mind, I know how Google works.



This is giving a lot of compiler errors like

asylvest: You use an unnamed namespace to basically hide functions in a cpp so that there's no way someone could link and use it.  So helper functions that aren't intended to be used elsewhere.  In an actual application cpp where nothing else is going to link in, it's just a style/consistency thing.

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
JonathanMeans: Would it work to consistently used six:UByte instead? I’m kind of worried about getting weird errors if getNumBytesPerPixel() is other than 2.

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
JonathanMeans: I think this will be a lot easier for me to add in once I’ve figured out what’s going on with the NITFReadControl. Other changes are made.

six/modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -251,6 +251,14 @@ class Utilities
     * \return the unit vector normal to the ground plane
     */
     static Vector3 getGroundPlaneNormal(const ComplexData& data);
+
+    /*
+     * Read a Data object from a SICD XML file.
+     * \param xmlPath The path to the input XML file
```
asylvest: Update comment - `xmlPathname`

six/modules/c++/six/source/Utilities.cpp
```diff
@@ -1198,3 +1199,20 @@ void six::getErrors(const ErrorStatistics* errorStats,
         }
     }
 }
+
+std::auto_ptr<Data> six::readXML(const std::string& xmlPath, DataType dataType)
```
asylvest: Should be `xmlPathname`

six/modules/c++/six/source/Utilities.cpp
```diff
@@ -1198,3 +1199,20 @@ void six::getErrors(const ErrorStatistics* errorStats,
...
+{
+    io::FileInputStream xmlFileStream(xmlPath);
+    xml::lite::MinidomParser treeBuilder;
+    treeBuilder.parse(xmlFileStream);
+    xmlFileStream.close();
+
+    std::auto_ptr<logging::Logger> log(new logging::NullLogger());
```
asylvest: Don't need to bother with a pointer here.  Just make
`logging::NullLogger logger`
and then send in `&logger`

six/modules/c++/six/source/Utilities.cpp
```diff
@@ -1198,3 +1199,20 @@ void six::getErrors(const ErrorStatistics* errorStats,
...
+    six::XMLControl *control = six::XMLControlFactory::getInstance().
+            newXMLControl(dataType, log.get());
+
+    std::auto_ptr<Data> data(control->fromXML(treeBuilder.getDocument(),
+        std::vector<std::string>()));
```
asylvest: Would be good to allow the caller to optionally pass in schemas (we want them too).  How about an optional parameter that defaults to an empty vector?

six/modules/c++/six/source/Utilities.cpp
```diff
@@ -1198,3 +1199,20 @@ void six::getErrors(const ErrorStatistics* errorStats,
...
+
+std::auto_ptr<Data> six::readXML(const std::string& xmlPath, DataType dataType)
+{
+    io::FileInputStream xmlFileStream(xmlPath);
+    xml::lite::MinidomParser treeBuilder;
+    treeBuilder.parse(xmlFileStream);
+    xmlFileStream.close();
+
+    std::auto_ptr<logging::Logger> log(new logging::NullLogger());
+    six::XMLControl *control = six::XMLControlFactory::getInstance().
```
asylvest: Doesn't this leak?

six/modules/c++/six.sicd/source/Utilities.cpp
```diff
@@ -507,6 +507,24 @@ Vector3 Utilities::getGroundPlaneNormal(const ComplexData& data)
...
+        const std::string& xmlPathname)
+{
+    XMLControlFactory::getInstance().addCreator(
+        DataType::COMPLEX,
+        new XMLControlCreatorT<ComplexXMLControl>());
+
+    std::auto_ptr<ComplexData> complexData(reinterpret_cast<ComplexData*>(
+            six::readXML(xmlPathname, DataType::COMPLEX).release()));
+
+    if (complexData.get() == NULL)
```
asylvest: This will never happen (assuming `six::readXML()` doesn't have a bug where it returns a NULL `auto_ptr`).  `dynamic_cast` would give you a NULL pointer if it was the wrong type but not `reinterpret_cast`.  I think you can stick with `reinterpret_cast` and assume that `six::readXML()` did the right thing when you passed in `COMPLEX`.

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,256 @@
...
+    }
+    if (!str::endsWith(argv[1], ".xml"))
+    {
+        std::string message = "Extension suggests input may not be an XML file. "
+            "Expecting '*.xml'.";
+        throw except::Exception(Ctxt(message));
+    }
+}
+
+std::vector<sys::Int16_T> generateBandData(const six::sicd::ComplexData& data)
```
asylvest: This doesn't matter a ton b/c it's just a test program, but this will potentially result in a copy depending on how smart the compiler is (C++11-compiler will likely optimize this out).

six/modules/c++/six.sidd/include/six/sidd/Utilities.h
```diff
@@ -70,6 +70,13 @@ class Utilities
     static std::auto_ptr<scene::ProjectionModel>
     getProjectionModel(const DerivedData* data);
 
+    /*
+    * Read a Data object from a SIDD XML file.
+    * \param xmlPath The path to the input XML file
```
asylvest: `xmlPathname`

six/modules/c++/six.sidd/source/Utilities.cpp
```diff
@@ -493,5 +494,24 @@ Utilities::getProjectionModel(const DerivedData* data)
...
+std::auto_ptr<DerivedData> Utilities::readXML(const std::string& xmlPathname)
+{
+    XMLControlFactory::getInstance().addCreator(
+        DataType::DERIVED,
+        new XMLControlCreatorT<DerivedXMLControl>());
+
+    std::auto_ptr<DerivedData> derivedData(reinterpret_cast<DerivedData*>(
+            six::readXML(xmlPathname, DataType::DERIVED).release()));
+
+    if (derivedData.get() == NULL)
```
asylvest: Ditto - not needed

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
JonathanMeans: It looks like this gets optimized out for MSVC since 2005 and gcc since 2002 (v. 3.1.1). But should I make it an output parameter instead just in case?

asylvest: Fair enough - then ok to leave as-is.

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,272 @@
...
+    }
+}
+
+
+bool addingUnloadedSegmentWriterShouldThrow(const std::string& xmlPathname)
+{
+    std::cout << "Running addingUnLoadedSegmentWriterShouldThrow\n";
+    std::auto_ptr<six::sicd::ComplexData> data =
+            six::sicd::Utilities::parseDataFromFile(xmlPathname,
+            std::vector < std::string>(),
```
asylvest: Extra whitespace... should just be `std::vector<std::string>()`

six/modules/c++/six.sicd/tests/test_add_additional_des.cpp
```diff
@@ -0,0 +1,272 @@
...
+    nitf::DESegment des = record.newDataExtensionSegment();
+    des.getSubheader().getFilePartType().set("DE");
+    des.getSubheader().getTypeID().set("XML_DATA_CONTENT_005");
+    des.getSubheader().getVersion().set("01");
+    des.getSubheader().getSecurityClass().set("U");
+
+    mem::SharedPtr<nitf::SegmentWriter> segmentWriter(new nitf::SegmentWriter);
+    writer.addAdditionalDES(segmentWriter);
+
+    std::string name(std::tmpnam(NULL));
```
asylvest: What if the writer starts to write the file and then throws for some reason?  You won't ever call `std::remove()` here.  A simpler way to do this is to make an RAII object that at construction calls `tmpnam()`, at destruction calls `std::remove`, and has a method to give you the name of the file so you can pass it to `writer.save()`.

PR #78, opened 2016-06-02
Handling extra XML_DATA_CONTENT DESs when reading

asylvest:
@JonathanMeans

### DIFF
--- a/six/modules/c++/six.sicd/tests/test_read_sicd_with_extra_des.cpp
+++ b/six/modules/c++/six.sicd/tests/test_read_sicd_with_extra_des.cpp
@@ -41,7 +41,6 @@ private:
 
 TempFile::TempFile() :
     mName(std::tmpnam(NULL))
-    //mName("C:\\Users\\jmeans.MDAUS\\Documents\\code\\six-library\\extraStuff.nitf")
 {
 }
 
@@ -89,12 +88,12 @@ std::auto_ptr<TempFile> createNITFFromXML(const std::string& xmlPathname)
 {
     logging::Logger log;
     std::auto_ptr<six::sicd::ComplexData> data =
-        six::sicd::Utilities::parseDataFromFile(xmlPathname,
+            six::sicd::Utilities::parseDataFromFile(xmlPathname,
             std::vector<std::string>(),
             log);
 
     std::vector<six::UByte> bandData(
-        generateBandData(*data));
+            generateBandData(*data));
 
     six::Container container(six::DataType::COMPLEX);
     container.addData(dynamic_cast<six::Data*>(data.release()));
@@ -109,16 +108,20 @@ std::auto_ptr<TempFile> createNITFFromXML(const std::string& xmlPathname)
     des.getSubheader().getVersion().set("01");
     des.getSubheader().getSecurityClass().set("U");
 
-    //des.getSubheader().getSubheaderFieldsLength().set(
-    //        six::Constants::DES_USER_DEFINED_SUBHEADER_LENGTH);
+    static const char segmentData[] = "123456789ABCDEF0";
+    nitf::SegmentMemorySource sSource(segmentData, strlen(segmentData),
+            0, 0, true);
+    mem::SharedPtr<nitf::SegmentWriter> segmentWriter(new nitf::SegmentWriter);
+    segmentWriter->attachSource(sSource);
+    writer.addAdditionalDES(segmentWriter);
 
     //Wrong tag
-    //nitf::TRE usrHdr("PIAIMB", "PIAIMB");
-    //record.getHeader().getUserDefinedSection().appendTRE(usrHdr);
+    nitf::TRE usrHdr("PIAIMB", "PIAIMB");
+    record.getHeader().getUserDefinedSection().appendTRE(usrHdr);
 
     //Good tag, wrong DESSHSI
     nitf::TRE secondHdr(six::Constants::DES_USER_DEFINED_SUBHEADER_TAG,
-        six::Constants::DES_USER_DEFINED_SUBHEADER_ID);
+            six::Constants::DES_USER_DEFINED_SUBHEADER_ID);
     secondHdr.setField("DESCRC", "99999");
     secondHdr.setField("DESSHFT", "XML00000");
     secondHdr.setField("DESSHDT", "2016-06-01T00:00:00Z");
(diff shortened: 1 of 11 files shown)


six/modules/c++/six/source/Types.cpp
```diff
@@ -53,6 +53,11 @@ const char Constants::DES_USER_DEFINED_SUBHEADER_TAG[] = "XML_DATA_CONTENT";
 const char Constants::DES_USER_DEFINED_SUBHEADER_ID[] = "XML_DATA_CONTENT_773";
 const sys::Uint64_T Constants::DES_USER_DEFINED_SUBHEADER_LENGTH = 773;
 
+const char Constants::SICD_DESSHSI[] =
```
asylvest: Update `NITFWriteControl` to use these too.  You may need to tweak how you did this a bit in order for the writer to use them...

six/modules/c++/six/source/NITFReadControl.cpp
```diff
@@ -195,33 +195,42 @@ void NITFReadControl::load(nitf::IOInterface& ioInterface,
...
-                    ostr << "Expected a user-defined subheader size of "
-                         << Constants::DES_USER_DEFINED_SUBHEADER_LENGTH
-                         << " but got " << subheaderLength;
-                    throw except::Exception(Ctxt(ostr.str()));
+                    continue;
                 }
 
                 nitf::TRE tre = subheader.getSubheaderFields();
+                std::string field;
+                try
```
asylvest: Rather than a try/catch, use `tre.exists()` to check first.  Since it's going to be really rare that you see an `XML_DATA_CONTENT` without a DESSHSI, and it's not in an inner loop anyway, this isn't a performance thing but a style thing... basically that we want to reserve this for more exceptional situations so that when we do see a try/catch, we think there must be something really weird going on and look at it carefully.

asylvest: Oh I see... after taking a look at the SIDD 1.1 branch's `NITFReadControl::getDataType()`, I'll revise this a little.  Here's the problem... if you don't have your plugin path set, this'll throw which you'll catch and then you'll skip over a real DES.  In that case, you want to throw an exception similar to what the branch does telling them they need to set their plugin path.

I'm not sure now why I didn't use `exists()` on the SIDD 1.1 branch... does this not behave well in these scenarios (TRE without this field, TRE with this field w/o plugin path set, TRE with this field with plugin path set)?

But... if you get an XML_DATA_CONTENT of length 5, even with your plugin path set, you won't have `DESSHSI` in there.  So, need to account for that too.

Since this change is already going to make us depend on them setting their plugin path when reading a SICD, can you fold in the `NITFReadControl::getDataType()` from the SIDD 1.1 branch as part of this work too?  Try to see if you can combine the logic for the two cleanly, at least for the `DESSHSI` piece.

JonathanMeans: Looking at getDataType(nitf::Record& record) in sidd_1.1, line 95 it starts to loop through the record’s DESegments. It looks like the loop will always either return or throw in the first iteration. Is this by design?

six/modules/c++/six.sicd/tests/test_read_sicd_with_extra_des.cpp
```diff
@@ -0,0 +1,186 @@
...
+    TempFile();
+    ~TempFile();
+    std::string pathname() const;
+private:
+    std::string mName;
+};
+
+TempFile::TempFile() :
+    mName(std::tmpnam(NULL))
+    //mName("C:\\Users\\jmeans.MDAUS\\Documents\\code\\six-library\\extraStuff.nitf")
```
asylvest: Remove commented-out line

six/modules/c++/six/source/NITFReadControl.cpp
```diff
@@ -81,21 +81,112 @@ six::PixelType getPixelType(nitf::ImageSubheader& subheader)
...
 namespace six
 {
+DataType NITFReadControl::getDataType(nitf::Record& record)
+{
+    nitf::List des = record.getDataExtensions();
+    if (des.isEmpty())
+    {
+        return DataType::NOT_SET;
+    }
+    nitf::DESegment segment = (nitf::DESegment) des.getFirst().getData();
```
asylvest: Add a comment that the SICD spec guarantees that the first DES will be the SICD/SIDD DES

six/modules/c++/six/source/NITFReadControl.cpp
```diff
@@ -178,80 +269,39 @@ void NITFReadControl::load(nitf::IOInterface& ioInterface,
...
-
-        // SICD/SIDD 1.0 specify DESID as XML_DATA_CONTENT
-        // Older versions of the spec specified it as SICD_XML/SIDD_XML
-        // Here we'll accept any of these under the assumption that it's not
-        // such an old version of the spec that the XML layout itself has
-        // changed (if it did, XMLControl will end up throwing anyway)
-        if (desid == "XML_DATA_CONTENT" ||
-            desid == "SICD_XML" ||
-            desid == "SIDD_XML")
+        if (getDataType(segment) == DataType::NOT_SET)
```
asylvest: Look at the SIDD 1.1 branch... the FTITLE check on line 254ish needs to get changed over to this too

six/modules/c++/six/source/NITFReadControl.cpp
```diff
@@ -178,80 +269,39 @@ void NITFReadControl::load(nitf::IOInterface& ioInterface,
...
-            std::auto_ptr<Data> data(parseData(*mXMLRegistry,
-                                               ioAdapter,
-                                               dataType,
-                                               schemaPaths,
-                                               *mLog));
-            if (data.get() == NULL)
-            {
-                throw except::Exception(Ctxt("Unable to transform XML DES"));
-            }
+            continue;
```
asylvest: Just to be explicit, add a comment stating this'll skip over non-SICD/SIDD DESs

six/modules/c++/six.sicd/tests/test_read_sicd_with_extra_des.cpp
```diff
@@ -0,0 +1,221 @@
...
+        const std::string xmlPathname(argv[1]);
+
+        six::XMLControlFactory::getInstance().addCreator(
+            six::DataType::COMPLEX,
+            new six::XMLControlCreatorT<six::sicd::ComplexXMLControl>());
+
+        std::auto_ptr<TempFile> nitf = createNITFFromXML(xmlPathname);
+        six::NITFReadControl reader;
+        reader.load(nitf->pathname());
+        std::cout << "Test passed\n";
```
asylvest: I think this is a good test but one thing is that you're just verifying reading didn't error out... not checking that we actually read the right thing out.  So after you load(), if you can check that the `ComplexData` looks as expected, you should be good.

Note that you don't want to do just a string comparison of the XML because Linux and Windows will format things differently unless you have a different XML file for each platform.  But you should be able to use some of those `six::sicd::Utilities` to quickly go from XML string to `ComplexData` and then use the equality operator to compare.

asylvest:
(Note to myself as much as anything... the only thing remaining to do is to update the test to confirm that the SICD XML is still getting parsed properly)

six/modules/c++/six.sicd/tests/test_read_sicd_with_extra_des.cpp
JonathanMeans: This is done.

I ran into a confusing comment while I was fixing this. In ReadControl.h, line 100-103, it claims “This container is not allocated, nor deleted by this class.” NITFReadControl both allocates and deletes the container. Is this meant to mean that subclasses are not required to allocate or delete the container? Or should I just change the comment?

asylvest: I think the comment is intending to say that this class (`ReadControl`) doesn't do any memory managing of it (which is the case)... `NITFReadControl` itself of course does do the memory management as you say.  So I think this comment is ok... long-term I would like to turn `mContainer` into a smart pointer to avoid this altogether.

PR #85, opened 2016-07-09
No longer require NITF_PLUGIN_PATH

asylvest:
Changed code to link in XML_DATA_CONTENT statically.  The underlying NITFReadControl and NITFWriteControl check the singleton `PluginRegistry` to ensure the handler has been added.

The net effect of this is that you don't need to set NITF_PLUGIN_PATH or call `six::loadPluginDir()` to read or write SICDs or SIDDs.  The only reason you'd still need to do this is if you're handling files that contain additional TREs.


asylvest:
@JonathanMeans @chvink @clydestanfield

PR #86, opened 2016-07-11
Lut reposition

JonathanMeans:
Allow for reading and writing LookupTable in the NITF that's not in the SICD.

### DIFF
--- a/six/modules/c++/six.sidd/include/six/sidd/DerivedData.h
+++ b/six/modules/c++/six.sidd/include/six/sidd/DerivedData.h
@@ -294,7 +294,7 @@ struct DerivedData: public Data
         return productCreation->classification;
     }
 
-    virtual LUT* getDisplayLUT();
+    virtual mem::ScopedCopyablePtr<LUT>& getDisplayLUT();
 
     virtual std::string getVendorID() const
     {
--- a/six/modules/c++/six.sidd/include/six/sidd/DerivedData.h
+++ b/six/modules/c++/six.sidd/include/six/sidd/DerivedData.h
@@ -319,6 +319,7 @@ private:
     static const char VENDOR_ID[];
     virtual bool equalTo(const Data& rhs) const;
     std::string mVersion;
+    LUT* mNITFDisplayLUT;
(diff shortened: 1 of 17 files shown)


six/modules/c++/six.sidd/include/six/sidd/DerivedData.h
```diff
@@ -117,6 +117,11 @@ struct DerivedData: public Data
      */
     Annotations annotations;
 
+    /*
+     * (Optional) In SIDD 1.1, the LUT stored in the NITF
+     */
```
asylvest: Add a comment here about why it's not needed for SIDD 1.0

JonathanMeans: Comment added.

six/modules/c++/six/source/NITFReadControl.cpp
```diff
@@ -418,6 +418,21 @@ void NITFReadControl::load(nitf::IOInterface& ioInterface,
...
 
+        // SIDD 1.1 needs to read LUT directly from NITF
+        if (currentInfo->getData()->getDataType() == DataType::DERIVED &&
+            currentInfo->getData()->getVersion() == "1.1.0")
+        {
+            nitf::LookupTable nitfLut =
+                    subheader.getBandInfo(0).getLookupTable();
+
+            //getDisplayLUT will not return NULL for 1.1
+            currentInfo->getData()->getDisplayLUT().reset(
```
asylvest: So rather than relying/hoping that someone always does this and remembers to call `interleaveTable()`, what about just adding a `six::LUT` constructor that takes in a `nitf::LookupTable` directly and does the right thing for you?

asylvest:
(Comment for myself - after issue with adding a new `six::LUT` constructor is addressed, this is good to go)

six/modules/c++/six/source/NITFReadControl.cpp
JonathanMeans: This is done.

PR #90, opened 2016-07-15
Fix how CSM figures out an images type, to match how NITFReadControl now does it.

JonathanMeans:


six/projects/csm/source/SIXSensorModel.cpp
```diff
@@ -1110,5 +1110,62 @@ double SIXSensorModel::getCorrelationCoefficient(size_t cpGroupIndex,
...
+    }
+    else if (desid == Constants::DES_USER_DEFINED_SUBHEADER_TAG)
+    {
+        // Check whether DES is SICD/SIDD, or something of a different
+        // sort. Note that we need to check the subheader
+        // length first rather than calling tre.getCurrentSize() in
+        // case there is no subheader (in which case
+        // subheader.getSubheaderFields() will throw with a
+        // NITRO-specific message).
+        const std::string& subheader = des.subHeader();
```
asylvest: XML_DATA_CONTENT is statically linked in now so this can use that now.  But can't you take it a step further and use a method from NITFReadControl so that the logic is in one spot (might need to refactor it a bit)?

JonathanMeans: I’ve made the edits on this, fixTests, and currentPath in CODA-OSS.

six/projects/csm/source/SIXSensorModel.cpp
```diff
@@ -1110,5 +1111,21 @@ double SIXSensorModel::getCorrelationCoefficient(size_t cpGroupIndex,
...
     }
 }
+
+DataType SIXSensorModel::getDataType(const csm::Des& des)
+{
+    std::string desid = des.subHeader().substr(NITF_DE_SZ, NITF_DESTAG_SZ);
+    str::trim(desid);
+    const std::string& subheader = des.subHeader();
+
+    const size_t desshsiOffset = 73 + 200;
```
asylvest: So it feels like logically you should be able to either reuse the `nitf_TREDescription` in `XML_DATA_CONTENT.c` and/or construct a `nitf::TRE` with `XML_DATA_CONTENT` and ask it for the field offset and length of this field.  Obviously in some sense the TRE class knows that.

It looks like you could do

```
nitf::TRE tre("XML_DATA_CONTENT");
const size_t desshsiLength = tre.getField("DESSHSI").getLength();
```

But I don't see a good way to get the offset... it would be nice if the `nitf::Field` knew that too.  You could probably iterate through all the TRE fields and add it up until you got there, but then I assume that 200 is the size of the DESubheader upfront so you'd still need that.

It doesn't feel good to have these three magic numbers in here but I don't think we can sanely get rid of all of them.  So let's do this...

- Use a named static constant for the 200 value (something like `DES_SUBHEADER_LENGTH` so you don't have that magic number in two spots
- See if the above code works to get the length of DESSHSI.  If so, use that rather than hardcoding 60
- Leave the 73 in there but with some comment that that's the offset to the start of DESSHSI

six/projects/csm/source/SIXSensorModel.cpp
```diff
@@ -1110,5 +1111,21 @@ double SIXSensorModel::getCorrelationCoefficient(size_t cpGroupIndex,
...
+    const size_t desshsiLength = 60;
+    std::string desshsiField = des.subHeader().substr(
+            desshsiOffset, desshsiLength);
+    str::trim(desshsiField);
+
+    return NITFReadControl::getDataType(desid, subheader.size() - 200,
```
asylvest: Well, since you already need to be dealing with this as a string anyhow... you can use the fourth argument and send in the DESID.  I think this should be bytes [2, 27] of the string.

six/modules/c++/six/source/NITFReadControl.cpp
```diff
@@ -127,34 +120,21 @@ DataType NITFReadControl::getDataType(nitf::DESegment& segment)
...
-
-        if (field == Constants::SICD_DESSHSI)
+        if (desshsiField == Constants::SICD_DESSHSI)
         {
             return DataType::COMPLEX;
         }
-        else if (field == Constants::SIDD_DESSHSI)
+        else if (desshsiField == Constants::SIDD_DESSHSI)
         {
             return DataType::DERIVED;
```
asylvest: Right now the function won't return anything if the conditions on lines 113, 117, and 121 are all false.  So, instead of the final else that returns `NOT_SET` on line 143, that should move all the way out and be the default return statement.

six/projects/csm/source/SIXSensorModel.cpp
asylvest: Actually... the other thing is that this won't work for SICDs and SIDDs before 1.x (that used SICD_XML / SIDD_XML for their DESID) right?  It's assuming it's grabbing one that's XML_DATA_CONTENT isn't it?

six/modules/c++/six.sicd/tests/test_read_sicd_from_vts.cpp
```diff
@@ -0,0 +1,321 @@
+/* =========================================================================
```
asylvest: Can you add something to the automated tests to call this?

six/modules/c++/six.sicd/tests/test_read_sicd_from_vts.cpp
```diff
@@ -0,0 +1,321 @@
...
+        TestScript testScript(testFile);
+        testScript.setSource(sicdPathname);
+        std::string vts = sys::Path::joinPaths("install",
+                sys::Path::joinPaths("bin", "vts"));
+        sys::OS os;
+        if (!os.isFile(vts))
+        {
+            vts = vts + ".exe";
+        }
+
```
asylvest: Can you add tests for SICD 0.4 (that'll have SICD_XML DESID)?

Also a test for SIDD 1.0 (don't actually worry about SIDD_XML DESID... that was some really old version of SIDD that's not supported).  We do want to make sure SIDD 1.0 works though.

six/modules/c++/six.sicd/tests/test_read_sicd_from_vts.cpp
asylvest: If calling vts is a pain, I wonder if it would be better to just instantiate the CSM plugin directly yourself in a test program and then you could easily call any method you want instead of jumping through weird vts hoops.  You could use `sys::DLL` to open the plugin.

JonathanMeans: So I’m not really sure how to set this up. When I had the test one of the generated files,  it looks like VTS tries to do an extra layer of validation and fails. Is there someplace I could put some real SICDs for Jenkins to pull from for running this?

six/projects/csm/source/SIXSensorModel.cpp
JonathanMeans: So what’s going on here is that NITFReadControl::getDataType returns right away if the DESID is SICD_XML or SIDD_XML. So grabbing the rest of the data assumes that it’s XML_DATA_CONTENT, since if it’s not, grabbing the wrong things won’t matter.

JonathanMeans: Does this end up being the same as the first argument, then? I went ahead and removed it completely for now, if this is the case.

six/modules/c++/six.sicd/tests/test_read_sicd_from_vts.cpp
JonathanMeans: I was able to find SICD 0.4.1 and SIDD 1.0.0 NITFs in V:ComSAR\, but I’m having trouble finding anything for SICD 1.0. Do you know where I could find one?

asylvest: Were you able to find a SICD 1.0?  If not, you can use one of the sample SICDs on the SIX wiki.

six/modules/c++/six/source/NITFReadControl.cpp
```diff
@@ -127,42 +120,60 @@ DataType NITFReadControl::getDataType(nitf::DESegment& segment)
...
+    std::string desid = subheader.getTypeID().toString();
+    str::trim(desid);
+
+    // Note that we need to check the subheader
+    // length first rather than calling tre.getCurrentSize() in
+    // case there is no subheader (in which case
+    // subheader.getSubheaderFields() will throw with a
+    // NITRO-specific message).
+    const sys::Uint64_T subheaderLength(subheader.getSubheaderFieldsLength());
+    std::string treTag("");
```
asylvest: Just default construct this and the next string... they'll be empty so no reason to do it explicitly.

six/projects/csm/source/SIXSensorModel.cpp
```diff
@@ -1110,5 +1111,38 @@ double SIXSensorModel::getCorrelationCoefficient(size_t cpGroupIndex,
...
+    }
+
+    const std::string& subheader = des.subHeader();
+
+    // DESSHSI occurs 73 bytes past the subheader
+    const size_t desshsiOffset = 73 + DES_SUBHEADER_LENGTH;
+
+    nitf::TRE tre("XML_DATA_CONTENT");
+    const size_t desshsiLength = tre.getField("DESSHSI").getLength();
+    std::string desshsiField = des.subHeader().substr(
```
asylvest: Can just use `subheader` rather than `des.subHeader()`

six/projects/csm/source/SIXSensorModel.cpp
```diff
@@ -1110,5 +1111,38 @@ double SIXSensorModel::getCorrelationCoefficient(size_t cpGroupIndex,
...
+            desshsiOffset, desshsiLength);
+    str::trim(desshsiField);
+
+    std::cerr << des.subHeader() << std::endl;
```
asylvest: Stray printout

six/modules/c++/six.sicd/tests/test_read_sicd_from_vts.cpp
JonathanMeans: I’ve found one, but VTS isn’t loading the plugin list for it. If the one on the wiki doesn’t work, I’ll try loading CSM directly, then try to figure out what’s going on there.

JonathanMeans: This is good to go now, running with SICD version 0.4.1 and 1.1.0, and SIDD version 1.1.0.

asylvest:
Looks good

PR #91, opened 2016-07-18
Sicd validation

JonathanMeans:


six/modules/c++/six.sicd/include/six/sicd/ComplexData.h
```diff
@@ -334,6 +334,9 @@ class ComplexData: public Data
     }
 
     bool operator==(const ComplexData& rhs) const;
+    bool validate(logging::Logger& log) const;
```
asylvest: Add Doxygen

asylvest: At least for now, just at this level... don't worry about doing it for the individual classes

six/modules/c++/six.sicd/include/six/sicd/Functor.h
```diff
@@ -0,0 +1,74 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef __SIX_SICD_VALIDATOR_H__
```
asylvest: Include guard should match filename

six/modules/c++/six.sicd/include/six/sicd/Functor.h
```diff
@@ -0,0 +1,74 @@
...
+public:
+    virtual std::vector<double> operator()(size_t n) const = 0;
+};
+
+class Identity : public Functor
+{
+public:
+    virtual std::vector<double> operator()(size_t n) const
+    {
+        (void)n;
```
asylvest: Comment out `n` above instead... just need to do the `(void)` bit for C code

six/modules/c++/six.sicd/include/six/sicd/Functor.h
```diff
@@ -0,0 +1,74 @@
...
+#define __SIX_SICD_VALIDATOR_H__
+
+#include "six/Init.h"
+#include <vector>
+
+namespace six
+{
+namespace sicd
+{
+class Functor
```
asylvest: Add some Doxygen for what this stuff is

six/modules/c++/six.sicd/include/six/sicd/Functor.h
```diff
@@ -0,0 +1,74 @@
...
+double bessi1(double x);
+double bessin(size_t order, double x);
+
+class Kaiser : public Functor
+{
+public:
+    Kaiser(double beta);
+    virtual std::vector<double> operator()(size_t n) const;
+private:
+    double beta;
```
asylvest: Should be `mBeta`

six/modules/c++/six.sicd/include/six/sicd/Functor.h
```diff
@@ -0,0 +1,74 @@
...
+class RaisedCos : public Functor
+{
+public:
+    RaisedCos(double coef);
+    virtual std::vector<double> operator()(size_t n) const;
+private:
+    double mCoef;
+};
+
+double bessi(size_t order, double x);
```
asylvest: What do you think about putting these in CODA-OSS's `math` module instead?

six/modules/c++/six.sicd/include/six/sicd/GeoData.h
```diff
@@ -27,12 +27,13 @@
...
 #include "six/ParameterCollection.h"
 #include <mem/ScopedCloneablePtr.h>
+#include <logging/Logger.h>
 
 namespace six
 {
 namespace sicd
 {
-
+struct ImageData;
```
asylvest: Throughout, include the needed headers rather than forward declaring.  That way, when someone includes this header, they don't need to also include other ones.

six/modules/c++/six.sicd/include/six/sicd/GeoData.h
```diff
@@ -98,6 +99,8 @@ class GeoData
     GeoData() :
         earthModel(EarthModelType::WGS84)
     {
+        scp.ecf = Init::undefined<Vector3>();
```
asylvest: Update SCP's constructor rather than doing that here

six/modules/c++/six.sicd/include/six/sicd/Grid.h
```diff
@@ -127,6 +141,42 @@ struct DirectionParameters
...
+            calculateImageVertices(const ImageData& imageData) const;
+
+    /* Return vector contents, in order:
+    * 0) deltaK1 (min)
+    * 1) deltaK2 (max)
+    */
+    std::pair<double, double> calculateDeltaKs(
+            const ImageData& imageData) const;
+
+    const double WGT_TOL = 1e-3;
```
asylvest: Here and throughout, make all of these `static`.  You'll end up needing to set the values themselves in the cpp file.

six/modules/c++/six.sicd/include/six/sicd/Grid.h
```diff
@@ -127,6 +141,42 @@ struct DirectionParameters
...
+    const size_t DEFAULT_WEIGHT_SIZE = 512;
+    const std::string boundsErrorMessage =
```
asylvest: Here and throughout, this should be all caps with underscores.  Also change it to a `const char []` to be slightly more efficient unless it makes usage more of a pain.

six/modules/c++/six.sicd/source/ComplexData.cpp
```diff
@@ -141,5 +141,166 @@ bool ComplexData::equalTo(const Data& rhs) const
...
+            grid->fillDerivedFields(*rma, geoData->scp.ecf, position->arpPoly);
+        }
+    }
+
+    if (includeDefault)
+    {
+        fillDefaultFields();
+    }
+
+    return;
```
asylvest: Not needed

asylvest:
Also, update `check_valid_six` to do the validation.  And add a Python equivalent script that does the same thing.

six/modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -253,7 +253,35 @@ class Utilities
...
+     * \return Poly1D in specified dimension
+     */
+    static Poly1D nPoly(PolyXYZ poly, size_t index);
+
+    /*
+     * Find sign of input, expressed as -1, 0, or 1
+     * \param val A signed number
+     * \return 1 if val is positive, -1 if negative, 0 if 0
+     */
+    template <class T>
```
asylvest: Use `typename` rather than `class`
I think this belongs in the `math` module too
Also, since you know `T` is POD, just pass it in by value

six/modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -253,7 +253,35 @@ class Utilities
...
+            return -1;
+        }
+        if (val > 0)
+        {
+            return 1;
+        }
+        return 0;
+    }
+
+    static Vector3 wgs84Norm(const Vector3& point);
```
asylvest: Need Doxygen

six/modules/c++/six.sicd/source/ComplexData.cpp
```diff
@@ -141,5 +141,166 @@ bool ComplexData::equalTo(const Data& rhs) const
...
+    // This function is a transcription of MATLAB file validate_sicd.m by Wade Schwartzkopf
+    // Reference numbers (e.g. 2.3) reference the corresponding sections of the MATLAB file
+    bool valid = (grid->validate(*collectionInformation, *imageData, log) &&
+            position->validate(log) &&
+            scpcoa->validate(*geoData, *grid, *position, log) &&
+            imageData->validate(*geoData, log) &&
+            geoData->validate(log) &&
+            radarCollection->validate(log));
+
+    double fc(Init::undefined<double>());
```
asylvest: This logic is repeated several times... make this a private member function

six/modules/c++/six.sicd/source/Functor.cpp
```diff
@@ -0,0 +1,211 @@
...
+        ret.push_back(1);
+        return ret;
+    }
+
+    size_t m = L - 1;
+    double k;
+    for (size_t ii = 0; ii < L; ++ii)
+    {
+        k = 2 * beta / m * std::sqrt(ii * (m - ii));
+        //k = beta * std::sqrt(1 - std::pow((2 * ii / L), 2));
```
asylvest: Commented out line is unneeded?

six/modules/c++/six.sicd/source/GeoData.cpp
```diff
@@ -49,6 +50,68 @@ bool GeoData::operator==(const GeoData& rhs) const
...
+        cornerLineSample.resize(4);
+        cornerLineSample[0].row = 1;
+        cornerLineSample[0].col = 1;
+        cornerLineSample[1].row = 1;
+        cornerLineSample[1].col = imageData.numCols;
+        cornerLineSample[2].row = imageData.numRows;
+        cornerLineSample[2].col = imageData.numCols;
+        cornerLineSample[3].row = imageData.numRows;
+        cornerLineSample[4].col = 1;
+        // line 746; requires point_slant_to_ground
```
asylvest: Can you use `scene::ProjectionModel::imageToScene()` for this?

six/modules/c++/six.sicd/source/GeoData.cpp
```diff
@@ -49,6 +50,68 @@ bool GeoData::operator==(const GeoData& rhs) const
...
+    }
+
+    // Derived: Add ValidData geocoords
+    if (!imageData.validData.empty() &&
+        validData.empty())
+    {
+        // line 757; requires point slant to ground
```
asylvest: Can you use `scene::ProjectionModel::imageToScene()` for this?

six/modules/c++/six.sicd/source/GeoData.cpp
```diff
@@ -49,6 +50,68 @@ bool GeoData::operator==(const GeoData& rhs) const
...
+
+bool GeoData::validate(logging::Logger& log) const
+{
+    bool valid = true;
+    std::ostringstream messageBuilder;
+
+    // 2.10
+    scene::LLAToECEFTransform transformer;
+    Vector3 derivedEcf = transformer.transform(scp.llh);
+    double ecf_diff = (scp.ecf - derivedEcf).norm();
```
asylvest: `ecfDiff` rather than `ecf_diff`

six/modules/c++/six.sicd/source/GeoData.cpp
```diff
@@ -49,6 +50,68 @@ bool GeoData::operator==(const GeoData& rhs) const
...
+
+    if (ecf_diff > 1e-2)
```
asylvest: Is there a reason this value isn't a private constant like the others?

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -62,6 +71,384 @@ bool DirectionParameters::operator==(const DirectionParameters& rhs) const
...
+            if (weightType->parameters.empty() || weightType->parameters[0].str().empty())
+            {
+                //A Hamming window is defined in many places as a raised cosine of weight .54,
+                //so this is the default. However, some data use a generalized raised cosine and
+                //call it HAMMING, so we allow for both uses.
+                coef = .54;
+            }
+            else
+            {
+                coef = str::toType<double>(weightType->parameters[0].str());
```
asylvest: In this and at least one other similar spot below, doesn't it work to just do...
`coef = weightType->parameters[0];`?
The `operator double()` (which I don't necessarily love but it's intended for stuff like this) should just do this for you...

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -62,6 +71,384 @@ bool DirectionParameters::operator==(const DirectionParameters& rhs) const
...
+
+    return weightFunction;
+}
+
+std::vector<RowColInt>
+DirectionParameters::calculateImageVertices(const ImageData& imageData) const
+{
+    std::vector<RowColInt> vertices;
+
+    if (imageData.validData.size() != 0)
```
asylvest: Use `!.empty()` rather than `.size() != 0`

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -62,6 +71,384 @@ bool DirectionParameters::operator==(const DirectionParameters& rhs) const
...
+        !Init::isUndefined<double>(kCenter))
+    {
+        deltaKCOAPoly = derivedKcoaPoly(rgAzComp, offset);
+    }
+}
+
+double DirectionParameters::derivedKCenter(const RgAzComp& rgAzComp,
+        double offset) const
+{
+    (void)rgAzComp;
```
asylvest: Just do `const RgAzComp& /*rgAzComp*/,` above rather than the `(void)`

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -62,6 +71,384 @@ bool DirectionParameters::operator==(const DirectionParameters& rhs) const
...
+    {
+        derivedCenter -= deltaKCOAPoly[0][0];
+    }
+    return derivedCenter;
+}
+
+Poly2D DirectionParameters::derivedKcoaPoly(const RgAzComp& rgAzComp,
+        double offset) const
+{
+    (void)rgAzComp;
```
asylvest: Ditto here and anywhere else there is a `(void)`

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -84,4 +471,831 @@ bool Grid::operator==(const Grid& rhs) const
...
+    const RadarModeType& mode = collectionInformation.radarMode;
+
+    //2.1. Scalar TimeCOAPoly means SPOTLIGHT data
+    bool isScalar = true;
+    bool valid = true;
+
+    // I don't know that it's impossible for a one-degree polynomial to be expressed
+    // as a polynomial of higher order for whatever reason, so I'm checking each term
+    // manually
+    for (size_t ii = 0; ii <= timeCOAPoly.orderX(); ++ii)
```
asylvest: This seems like a good thing to make a member function of the `TwoD` class... maybe `isScalar()`?

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -84,4 +471,831 @@ bool Grid::operator==(const Grid& rhs) const
...
+    {
+        row->unitVector = derivedRowUnitVector(rmcr, scp);
+        col->unitVector = derivedColUnitVector(rmcr, scp);
+    }
+}
+
+void Grid::fillDerivedFields(const INCA& inca, const Vector3& scp,
+        const PolyXYZ& arpPoly)
+{
+
```
asylvest: Extra newline

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -84,4 +471,831 @@ bool Grid::operator==(const Grid& rhs) const
...
+    {
+        row->unitVector = derivedRowUnitVector(scpcoa, scp);
+    }
+    if (Init::isUndefined<Vector3>(col->unitVector))
+    {
+        col->unitVector = derivedColUnitVector(scpcoa, scp);
+    }
+
+    row->fillDerivedFields(rgAzComp, geoData,
+            fc * 2 / math::Constants::SPEED_OF_LIGHT_METERS_PER_SEC);
```
asylvest: In the calling logic above, `fc` could be set to the reserved undefined value right?  So here it'll get set to some weird non-undefined value.  Does the Matlab code handle this any differently?

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -84,4 +471,831 @@ bool Grid::operator==(const Grid& rhs) const
...
+        else
+        {
+            Poly2D differencePoly = kcoaPoly - (centroidPoly * inca.timeCAPoly[1]);
+            double norm = 0;
+
+            for (size_t ii = 0; ii < differencePoly.orderX(); ++ii)
+            {
+                for (size_t jj = 0; jj < differencePoly.orderY(); ++jj)
+                {
+                    norm += std::pow(differencePoly[ii][jj], 2);
```
asylvest: Here and all the other spots that you use `std::pow(someValue, 2)`... `std::pow()` is generally dumbly inefficient in cases like this.  It is a lot more efficient to have...

```
inline double square(double val)
{
    return val * val;
}
```

In fact, if you grep through the SIX repo, you'll see that we've done that in a few cpp files.  Go ahead and add this function into the `math` module too and then replace all the uses of it in SIX as well as the `std::pow(someValue, 2)` cases you've added with it.

Not that efficiency really matters here, but it's no more complex to use this instead so we might as well.

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -84,4 +471,831 @@ bool Grid::operator==(const Grid& rhs) const
...
 
+ComplexImagePlaneType Grid::defaultPlaneType(const RMA& rma) const
+{
+    if (rma.rmat.get() || rma.rmcr.get())
+    {
+        return ComplexImagePlaneType::SLANT;
+    }
+    else if (rma.inca.get())
+    {
+        return this->imagePlane;
```
asylvest: Is `this->` needed here?

six/modules/c++/six.sicd/source/ImageData.cpp
```diff
@@ -36,4 +37,81 @@ bool ImageData::operator==(const ImageData& rhs) const
...
+            {
+                messageBuilder << "ImageData.ValidData first col of matching"
+                    << "minimum row index should have minimum col index";
+                log.error(messageBuilder.str());
+                valid = false;
+            }
+        }
+    }
+
+    //TODO: figure out how to tell if corners are clockwise
```
asylvest: Can you look at the SICD spec and see if it really requires the corners to be in clockwise order or for the check above in terms of the order of the corners to hold?  If not, let's talk.  If it does, we can talk about options for getting the corners in clockwise order.

six/modules/c++/six.sicd/source/SCPCOA.cpp
```diff
@@ -59,3 +67,314 @@ bool SCPCOA::operator==(const SCPCOA& rhs) const
...
+
+    // Project north onto ground plane
+    Vector3 northGround = base -
+            (earthTangentPlane(scp) * earthTangentPlane(scp).dot(base));
+    return northGround.unit();
+}
+
+double SCPCOA::derivedTwistAngle(const Vector3& scp) const
+{
+    // Angle from +GPY axis to the +SPY axis in plane of incidence
```
asylvest: For this and pretty much all of these derived angle methods... can you get them from a `SceneGeometry` object?  See `six::sicd::Utilities::getSceneGeometry()` for how to make one.  If it's not obvious how to do this cleanly because you've just got an `SCPCOA` object, let's talk.

six/modules/c++/six.sicd/source/SCPCOA.cpp
```diff
@@ -59,3 +67,314 @@ bool SCPCOA::operator==(const SCPCOA& rhs) const
...
+        vectors[0][ii] = posPoly(scpTime);
+        vectors[1][ii] = velPoly(scpTime);
+        vectors[2][ii] = accPoly(scpTime);
+    }
+    return vectors;
+}
+
+SideOfTrackType SCPCOA::derivedSideOfTrack(const GeoData& geoData) const
+{
+    return look(geoData.scp.ecf) < 0 ? SideOfTrackType::RIGHT :
```
asylvest: Ditto... can you get this from `SceneGeometry`?

six/modules/c++/six.sicd/source/SCPCOA.cpp
```diff
@@ -59,3 +67,314 @@ bool SCPCOA::operator==(const SCPCOA& rhs) const
...
+        messageBuilder << mScpcoaInconsistentString << "\n" <<
+            "SCPCOA.scpTime: " << scpTime << std::endl <<
+            "Derived scpTime: " << derivedSCPTime(grid) << std::endl;
+        log.error(messageBuilder.str());
+        valid = false;
+    }
+
+    std::vector<Vector3> arpVectors = derivedArpVectors(position);
+    // 2.7.2
+    valid = valid && compareFields(arpPos, arpVectors[0], "arpPos", log);
```
asylvest: OK so this may suck in terms of a refactor it would require, but it just occurred to me... if we want to validate a SICD, we probably want the validator to tell us ALL the stuff that's wrong with it.  The way checks like this are structured, if `valid` is false by the time we get here, `compareFields()` isn't going to get called because the first half of the conditional is already false.  One way to refactor would be to have some of these helper functions take in `valid` by non-const ref and update it.  For convenience, they could also return the value.  That is, it could be something like...

```
bool myComparison(int whateverOtherValues, bool& valid)
{
    bool myCompIsValid = true;

    if (whateverCondition)
    {
        myCompIsValid = false;
    }

    valid &=myCompIsValid;
    return myCompIsValid;
}
```

six/modules/c++/six.sicd/source/SCPCOA.cpp
asylvest: If it doesn't make sense to construct a `SceneGeometry` object, but the underlying math is the same, another option would be to make public static member functions of `SceneGeometry` that take in several parameters and call that version here (and have the existing `SceneGeometry` member functions call the new functions with their member variables)

six/modules/c++/six.sicd/unittests/test_filling_geo_data.cpp
```diff
@@ -0,0 +1,49 @@
...
+        // MATLAB give.
+        TEST_ASSERT_LESSER(std::abs(expected[ii] - geoData.scp.ecf[ii]), 100);
+    }
+}
+
+int main(int, char**)
+{
+    TEST_CHECK(ecfFromLlh);
+    return 0;
+}
```
asylvest: This and several of the other unittest files need a newline at the end (Unix wants another newline in general or else commands like `more` look funky in the shell)

six/modules/c++/six.sicd/unittests/test_filling_grid.cpp
```diff
@@ -0,0 +1,340 @@
...
+    row.weightType.reset(new six::sicd::WeightType());
+    row.weightType->windowName = "HAMMING";
+
+    six::sicd::ImageData imageData;
+    row.fillDerivedFields(imageData);
+    TEST_ASSERT_EQ(row.weights.size(), 512);
+    //Just grabbing some random values to make sure they line up with MATLAB
+    //TEST_ASSERT_ALMOST_EQ is too picky for what we have, so doing something
+    //more manual.
+    TEST_ASSERT_LESSER(std::abs(row.weights[163] - .7332), .0001);
```
asylvest: For this and the others, can you use `TEST_ASSERT_ALMOST_EQ_EPS()`?

six/modules/c++/six.sicd/include/six/sicd/GeoData.h
JonathanMeans: I’m having dependency issues when trying to include Init.h in Types.h, since the former relies on a lot of things the latter defines. Should I just set it manually to the value? Or I could start a Types.cpp.

There were similar issues with including struct header files in each other, vs. forward declaring, though I don’t recall the specifics. I can at least try to include what I can there.

JonathanMeans:
This is ready to go again

six/modules/c++/six.sicd/include/six/sicd/Functor.h
```diff
@@ -0,0 +1,76 @@
...
+ *   \brief Represents a windowing function, with set parameters
+ *
+ *   This class is meant as an interface for classes to represent windowing
+ *   functions. The parameters are set in the constructor, and the functor
+ *   is called with the window size
+ */
+class Functor
+{
+public:
+    virtual std::vector<double> operator()(size_t n) const = 0;
```
asylvest: Add in a virtual destructor here to be safe

six/modules/c++/six.sicd/source/ComplexData.cpp
```diff
@@ -141,5 +141,169 @@ bool ComplexData::equalTo(const Data& rhs) const
...
+        {
+            rgAzComp->fillDerivedFields(*geoData, *grid, *scpcoa, *timeline);
+            grid->fillDerivedFields(*rgAzComp, *geoData, *scpcoa, fc);
+        }
+        break;
+    case ImageFormationType::PFA:
+        if (pfa.get())
+        {
+            pfa->fillDerivedFields(*position);
+        }
```
asylvest: Want a `break` here right?

six/modules/c++/six.sicd/source/ComplexData.cpp
```diff
@@ -141,5 +141,169 @@ bool ComplexData::equalTo(const Data& rhs) const
...
+    case ImageFormationType::RMA:
+        if (rma.get())
+        {
+            rma->fillDerivedFields(*geoData, *position);
+            grid->fillDerivedFields(*rma, geoData->scp.ecf, position->arpPoly);
+        }
```
asylvest: Want a `break` here

asylvest: Is gcc or VS giving warnings about not having a `default` case in here?  If so, can you add a
```
default:
    break;
```

Here and anywhere else you're getting a warning?  If you're not getting warnings for this, don't worry about it.

six/modules/c++/six.sicd/source/GeoData.cpp
```diff
@@ -19,36 +19,113 @@
...
+    return new GeoInfo(*this);
+}
+
+bool GeoInfo::operator==(const GeoInfo& rhs) const
+{
+    return (name == rhs.name && geoInfos == rhs.geoInfos &&
+            desc == rhs.desc && geometryLatLon == rhs.geometryLatLon);
+}
+
+const double GeoData::ECF_THRESHOLD = 1e-2;
```
asylvest: For consistency with other files, put class constants like this one above all the other methods

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -62,6 +77,381 @@ bool DirectionParameters::operator==(const DirectionParameters& rhs) const
...
+    }
+    return;
+}
+
+bool DirectionParameters::validate(const ImageData& imageData,
+    logging::Logger& log) const
+{
+    bool valid = true;
+    std::ostringstream messageBuilder;
+    const double& epsilon = std::numeric_limits<double>::epsilon();
```
asylvest: I don't think it'll matter much performance-wise either way, but in general, making a reference to a POD item like this isn't going to save you anything (`sizeof(double&)` == `sizeof(double)`), so I'd say just hold onto a `const double` here.

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -72,6 +462,12 @@ Grid::Grid() :
 {
 }
 
+const double Grid::UVECT_TOL = 1e-3;
```
asylvest: Move these to the top of the file too

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -84,4 +480,812 @@ bool Grid::operator==(const Grid& rhs) const
...
+    if (imagePlane == ComplexImagePlaneType::NOT_SET)
+    {
+        imagePlane = ComplexImagePlaneType::SLANT;
+    }
+    if (imagePlane == ComplexImageGridType::NOT_SET)
+    {
+        type = ComplexImageGridType::RGAZIM;
+    }
+
+    if (Init::isUndefined<Vector3>(row->unitVector))
```
asylvest: Do you really need the explicit `<Vector3>` here and in the line below?  I would've thought the compiler would deduce it.

six/modules/c++/six.sicd/source/Grid.cpp
```diff
@@ -84,4 +480,812 @@ bool Grid::operator==(const Grid& rhs) const
...
+            << defaultGridType(rma).toString() << ".\nFound " << type;
+        log.error(messageBuilder.str());
+        valid = false;
+    }
+
+    if (rma.rmat.get())
+    {
+        return validate(*rma.rmat, scp, fc, log);
+    }
+
```
asylvest: In chained `if / else if` spots like this, don't have a newline between the blocks

six/modules/c++/six.sicd/source/RadarCollection.cpp
```diff
@@ -70,6 +76,174 @@ WaveformParameters* WaveformParameters::clone() const
...
+    if (rcvDemodType == DemodType::STRETCH &&
+        std::abs(rcvFMRate / txFMRate - 1) > WGT_TOL)
+    {
+        messageBuilder.str("");
+        messageBuilder << WF_INCONSISTENT_STR << std::endl
+            << "SICD.RadarCollection.Waveform.WFParameters.RcvDemodType: "
+            << rcvDemodType << std::endl
+            << "SICD.RadarCollection.Waveform.WFParameters.RcvFMRate: "
+            << rcvFMRate << std::endl
+            << "SICD>RadarCollection.Waveform.WFParameters.TxFMRate: "
```
asylvest: Typo: `SICD>RadarCollection` should be `SICD.RadarCollection`

six/modules/c++/six.sicd/source/RadarCollection.cpp
```diff
@@ -171,6 +344,10 @@ Area* Area::clone() const
     return new Area(*this);
 }
 
+const double RadarCollection::WF_TOL = 1e-3;
```
asylvest: Move to top of file

six/modules/c++/six.sicd/source/RgAzComp.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include "six/sicd/SCPCOA.h"
+#include "six/sicd/Timeline.h"
+#include "six/sicd/Utilities.h"
+#include <math/Utilities.h>
+
+namespace six
+{
+namespace sicd
+{
+
```
asylvest: Extra newlines

six/modules/c++/six.sicd/source/SCPCOA.cpp
```diff
@@ -59,3 +68,236 @@ bool SCPCOA::operator==(const SCPCOA& rhs) const
...
+        {
+            arpVel = vectors[1];
+        }
+        if (Init::isUndefined<Vector3>(arpAcc))
+        {
+            arpAcc = vectors[2];
+        }
+    }
+
+    scene::SceneGeometry geometry(arpVel, arpPos, geoData.scp.ecf);
```
asylvest: Can be `const`

six/modules/c++/six.sicd/source/SCPCOA.cpp
```diff
@@ -59,3 +68,236 @@ bool SCPCOA::operator==(const SCPCOA& rhs) const
...
+    {
+        groundRange = derivedGroundRange(scp);
+    }
+    if (Init::isUndefined<double>(dopplerConeAngle))
+    {
+        dopplerConeAngle = geometry.getDopplerConeAngle();
+    }
+    if (Init::isUndefined<double>(grazeAngle))
+    {
+        grazeAngle = std::fabs(geometry.getETPGrazingAngle());
```
asylvest: Just use `std::abs()` rather than `std::fabs()` here and below - it's overloaded so that it'll do what you want for a `double`

six/modules/c++/six.sicd/source/Utilities.cpp
```diff
@@ -509,6 +510,64 @@ Vector3 Utilities::getGroundPlaneNormal(const ComplexData& data)
...
+    {
+        coefs.push_back(poly[ii][idx]);
+    }
+    Poly1D ret(coefs);
+    return ret;
+}
+
+Vector3 Utilities::wgs84Norm(const Vector3& point)
+{
+    scene::WGS84EllipsoidModel model;
```
asylvest: It seems like this should be added as a new method of `EllipsoidModel` instead of being in here... any reason not to?

six/modules/python/six.sicd/tests/validate_sicd.py
```diff
@@ -0,0 +1,60 @@
...
+from coda.xml_lite import *
+from coda.coda_io import *
+from coda.coda_logging import *
+
+import os
+import sys
+import os.path
+import filecmp
+
+##############################################################
```
asylvest: Update the comments to match what this is really doing.  Probably rename the function to `validateSicd()`.  Would be nice if this script returned 0 or 1 based on if all the SICDs you sent in validated.

six/modules/python/six.sicd/tests/validate_sicd.py
```diff
@@ -0,0 +1,60 @@
...
+# be any in the repository, so this is really more of an example.
+
+def loadSicd(filename):
+  vs = VectorString()
+  vs.push_back( os.environ['SIX_SCHEMA_PATH'] )
+  cmplx = SixSicdUtilities.getComplexData(filename, vs)
+  return cmplx.validate(getLogger())
+
+if 1 == len(sys.argv):
+    print("please provide a sicd or XML file")
```
asylvest: The comment says we support XML files but then we don't look for them below

six/modules/python/six.sicd/tests/validate_sicd.py
```diff
@@ -0,0 +1,60 @@
...
+from pysix.six_sicd import *
+from pysix.six_base import *
+from coda.xml_lite import *
+from coda.coda_io import *
+from coda.coda_logging import *
+
+import os
+import sys
+import os.path
+import filecmp
```
asylvest: Not all of these imports are needed

six/modules/c++/samples/check_valid_six.cpp
```diff
@@ -148,16 +148,26 @@ int main(int argc, char** argv)
             log->info(Ctxt("Reading " + inputPathname));
             try
             {
-                if (nitf::Reader::getNITFVersion(inputPathname) ==
-                        NITF_VER_UNKNOWN)
+                std::auto_ptr<six::Data> data =
```
asylvest: This doesn't work right if I send in a NITF - `parseDataFromFile()` is counting on getting in an XML text file

six/modules/c++/samples/check_valid_six.cpp
```diff
@@ -148,16 +148,26 @@ int main(int argc, char** argv)
...
+                        six::parseDataFromFile(xmlRegistry,
+                                               inputPathname,
+                                               schemaPaths,
+                                               *log);
+
+                six::sicd::ComplexData* complexData =
```
asylvest: Rather than a `dynamic_cast`, you can use `six::Data::getDataType()` and see if it is complex.  If it is, you can use `reinterpret_cast`.  This is cheaper (`dynamic_cast` needs to do an expensive check) - not that performance matters for this - and is consistent with how other parts of SIX do this check (which admittedly I have mixed feelings about).

six/modules/c++/six.sicd/source/Utilities.cpp
asylvest: Hmm... That's a good point and I don't know.  How about putting it just in `WGS84EllipsoidModel` to be safe then?

JonathanMeans:
This is ready again. The deleted function in DerivedXMLParser was a duplicated, probably from a bad merge.

asylvest:
This is close... these updates did introduce some compiler warnings at least on Windows.  Do `waf build --target=six-c++`.  Don't worry about any warnings up to this point.  Then do `waf build --target=six.sicd-c++` and fix at a minimum any warnings introduced from this pull request (I don't think it would take much to fix the other warnings at the same time so bonus points for that).  Most of these warnings are either about casts (make sure to use C++ and not C casts for updates here) and unused variables, but there is one about a declaration hiding a class member that definitely needs to be looked at.

JonathanMeans:
I've fixed the warnings.

asylvest:
Thanks.  I verified that I didn't see any bugs introduced (signed vs. unsigned kinds of issues or anything funky).  There were a couple style things I tweaked... figured it was easier just to commit these than to go through the thousandth round of reviews.

So... this is good to go!!

PR #95, opened 2016-08-02
Make crop_sicd's method of finding the appropriate directories more reliable.

JonathanMeans:


six/modules/python/six/tests/utils.py
```diff
@@ -43,6 +43,13 @@ def findSixHome():
 
     return currentPath
 
+def sicdDir():
```
asylvest: Can we put these directories in Jenkins rather than SIX?

JonathanMeans: This is accomplished.

PR #107, opened 2016-08-17
Don't get numBands from the size of bandDescriptors

JonathanMeans:
This will fail when the BandDescriptor elements are not included.

### DIFF
--- a/six/modules/c++/six.sidd/include/six/sidd/Display.h
+++ b/six/modules/c++/six.sidd/include/six/sidd/Display.h
@@ -206,6 +206,9 @@ struct BandInformation
 {
     BandInformation();
 
+    size_t numBands;
+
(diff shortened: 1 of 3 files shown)


six/modules/c++/six.sidd/include/six/sidd/Display.h
```diff
@@ -206,6 +206,9 @@ struct BandInformation
 {
     BandInformation();
 
+    size_t numBands;
```
asylvest: Can you initialize this in the constructor to the same reserved uninitialized value that `displayFlag` gets set to?

Other than that, looks good.

JonathanMeans: Done.

PR #110, opened 2016-09-02
Add setup.py for creating wheels

JonathanMeans:


setup.py
```diff
@@ -0,0 +1,48 @@
...
+
+codaPyds = glob.glob(os.path.join(packages, 'coda', '_*pyd'))
+codaPyds = [os.path.basename(pyd) for pyd in codaPyds]
+
+sixPyds = glob.glob(os.path.join(packages, 'pysix', '_*pyd'))
+sixPyds = [os.path.basename(pyd) for pyd in sixPyds]
+
+
+setup(name = 'pysix',
+      version = '2.2.1',
```
asylvest: From a synchronization perspective, I'd really like to have the version set in just one spot if possible so we don't forget to increment it (it used to be spread across multiple wscripts and was easy to get out of sync).  Logically you'd want it in the top-level SIX wscript... but the problem is that the way we subtree pull this into other repos, and the way waf works, we won't actually call the top-level wscript when we're in that other repo.  So for that reason, it's actually set in six/six/modules/wscript right now.

Any thoughts on how we can keep this to just one version?  If it's completely impossible / really problematic, we can do it as two...

JonathanMeans: This is ready for review again.

JonathanMeans: Ugh, never mind, something isn’t working with Python 3.3

JonathanMeans: It’s ready again.

asylvest:
The other wscripts that depend on the `VERSION` variable need to get updated to use `SIX_VERSION` now...  you want to change

```
modArgs['VERSION'] = bld.env['VERSION']
```

to
```
modArgs['VERSION'] = bld.env['SIX_VERSION']
```

JonathanMeans:
I’ve updated most of them. In six/modules/c++/samples/wscript, there’s  “VERSION = ‘0.1’” at the top. I’m not sure what that’s referencing, or if it needs to be changed at all.

asylvest:
Don't worry about that `VERSION` in the samples wscript... it's not technically used for anything.

asylvest:
So, some thoughts...
* This obviously predates your changes, but I just noticed in the top-level wscript that `distclean` is pointing at the wrong directories... it should use `DIRS`.  Can you fix this?
* I think the conf dir should move under the six subdir actually (parallel with modules and projects).  Otherwise, `waf install` doesn't seem like it does that target.  Plus this way the target goes to target/six/conf/setup.py so that some other setup.py down the road doesn't clobber it.
* Rename the target to 'six_setup'... 'setup' seems pretty generic and we want to make sure we don't clobber another target by the same name
* `waf package` is assuming that the target directory is named 'target', but it might not be.  At 'waf configure' time, someone can name this whatever they want via the `--out=` flag.  At build time, you can get this directory via `bld.bldnode.abspath()`.  More on this in a sec.
* `waf package` doesn't know where the installation directory is, so it's looking for a bin directory.  You'll probably be safe with this most of the time but of course someone could do something weird like put their own bin directory somewhere and this'll get fooled.  At build time, `bld.env['PREFIX']` is the installation directory (actually technically `bld.env['install_pydir']` is supposed to do this but I'm nearly positive it's busted right now so ignore that part for a sec).
* The problem is that at `waf package` time, you don't seem to have access to this stuff.  There may be another way to do this, so feel free to look at options, but the way we've added other commands like this in the past is to add a new file in the coda-oss/build directory and then some stuff in build.py at the bottom that inherits from our special context.  Look at `build/dumpconfig.py` and `CPPDumpConfigContext` for an example.  This would actually be good to put in CODA-OSS anyway because this is something we want to start being able to use in other repos.  dumpconfig is able to get at this kind of environment stuff.
* I noticed when it ran that it downloaded NumPy and I assume put it in the wheel.  This is because it could tell it was a dependency?  What does this do if you're not attached to a network?  Is there a way to at least optionally turn this off so we can have a smaller wheel?
* Can you make a separate command/option/something for generating the wheel vs. the zip file of the whole installation?

On the upside, the actual functionality we get for a wheel is exactly what we want; this is going to be a big improvement over how we do things now.

asylvest:
This works great!!

PR #114, opened 2016-09-27
Sidd comparison program

chvink:
This test program lets us compare two SIDDs to see if they're binary identical. It will optionally allow you to disable checking metadata, and optionally allow you to disable checking date related fields (eg creation date, processing times, etc).

This is mostly generalized for SICDs too, except `--ignore-date` hasn't been specialized for `six::sicd::ComplexData` yet.


six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
```diff
@@ -0,0 +1,208 @@
...
+            new six::XMLControlCreatorT<
+            six::sidd::DerivedXMLControl>()); 
+
+    six::NITFReadControl reader;
+    reader.setXMLControlRegistry(&xmlRegistry);
+
+    reader.load(sidd1Path);
+    mem::SharedPtr<six::Data> sidd1Metadata = readMetadata(reader);
+    mem::ScopedAlignedArray<sys::Int8_T> sidd1Buffer;
+    readWideband(reader,
```
asylvest: You could simplify this `readWideband()` function if you just took in a const ref to a `Data` object - then you could take in just one param for that rather than these three

six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
```diff
@@ -0,0 +1,208 @@
...
+    xmlRegistry.addCreator(six::DataType::DERIVED,
+            new six::XMLControlCreatorT<
+            six::sidd::DerivedXMLControl>()); 
+
+    six::NITFReadControl reader;
+    reader.setXMLControlRegistry(&xmlRegistry);
+
+    reader.load(sidd1Path);
+    mem::SharedPtr<six::Data> sidd1Metadata = readMetadata(reader);
+    mem::ScopedAlignedArray<sys::Int8_T> sidd1Buffer;
```
asylvest: Obviously it's one byte either way, but I'd say just stay consistent and use the `six::UByte` class for this since that's what SIX uses... then you don't need the cast in `setBuffer()` above either

six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
```diff
@@ -0,0 +1,208 @@
...
+        bool metadataMatches = (*sidd1Metadata) == (*sidd2Metadata);
+
+        if(!metadataMatches)
+        {
+            return false;
+        }
+    }
+    else
+    {
+        if (sidd1Metadata->getNumRows() != sidd2Metadata->getNumRows() ||
```
asylvest: Why is this in an `else` block?  Don't you want to always do this (otherwise the `memcmp()` could go out of bounds)?

six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    return !result;
+
+}
+
+int main(int argc, char* argv[])
+{
+    try 
+    {
+
```
asylvest: Extra newline

six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
```diff
@@ -0,0 +1,208 @@
...
+        if (siddsMatch(sidd1Path, sidd2Path, ignoreMetadata, ignoreDate))
+        {
+            return 0;
+        }
+        else
+        {
+            return 1;
+        }
+
```
asylvest: Extra newlines

six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    }
+    catch (except::Throwable & exception)
```
asylvest: Have the whole bottom of this look like round_trip_six - there are some extra things to catch

six/modules/c++/six.sidd/wscript
```diff
@@ -1,6 +1,7 @@
 NAME            = 'six.sidd'
 MAINTAINER      = 'adam.sylvester@mdaus.com'
 MODULE_DEPS     = 'scene tiff nitf xml.lite six mem'
+TEST_DEPS       = 'cli six.sicd'
```
asylvest: Rather than doing this and getting this cross thing where SIDD knows about SICD now (even though it's only in the test program), just move this into the modules/c++/samples folder that knows about both SICD and SIDD

six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include <import/mem.h>
+#include <import/except.h>
+
+#include <import/six.h>
+#include <import/six/sicd.h>
+#include <import/six/sidd.h>
+
+#include <cstring>
+
+mem::SharedPtr<six::Data> readMetadata(six::NITFReadControl& reader)
```
asylvest: Put all these functions in an unnamed namespace

six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
```diff
@@ -0,0 +1,208 @@
+#include <import/cli.h>
```
asylvest: This is just a pet peeve, but I hate these import headers.  I guess you don't have to, but bonus points if you change this to just bring in what you need from each module.

six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
asylvest: `reader` can come in by const ref right?

six/modules/c++/six.sidd/tests/test_compare_sidd.cpp
chvink: `(*sidd1Metadata) == (*sidd2Metadata)` should mean that the image dimensions match, I think the explicit check is only needed if we're ignoring metadata.

asylvest: Oh duh - yeah I agree

chvink:
All these issues should be addressed.

asylvest:
Looks good.  I made one little tweak where there's no reason to pass in `Data` as a shared pointer to `readWideband()`, so changed it to const ref (i.e. the idea is only pass smart pointers around when something needs to hold onto it).

PR #116, opened 2016-10-11
Install schemas with pysix and add accessor for schema path

tolsonMDA:
Add SICD schema path accessor to eliminate dependence on environment variable.


asylvest:
@JonathanMeans @chvink FYI

PR #122, opened 2016-11-01
io::SeekableInputStream Reader

clydestanfield:
This allows the six::NITFReadControl to read directly from an io::SeekableInputStream.

This depends on https://github.com/mdaus/nitro/pull/26 and should not be merged until that has been merged and subtree pulled into SIX.


asylvest:
Merged NITRO in - this change looks good too

Issue #123, opened 2016-11-02
Add test showing io::SeekableInputStream works with NITFReadControl


asylvest:
Add a test showing the new `six::NITFReadControl::load()` overloading that takes in an `io::SeekableInputStream` works (#122).  You can just turn around and create an `io::FileInputStream` and then call this method if you want.

@clydestanfield

PR #124, opened 2016-11-03
Test loading nitf from input stream and fix segfault

JonathanMeans:
@clydestanfield @mbojrab @asylvest #123

### DIFF
--- a/six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
+++ b/six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
@@ -28,6 +28,7 @@
 #include <sys/Path.h>
 #include <except/Exception.h>
 #include <io/StandardStreams.h>
+#include <six/Utilities.h>
 #include <six/sicd/ComplexXMLControl.h>
 #include <six/sicd/Utilities.h>
 #include <six/sicd/ComplexData.h>
@@ -55,12 +56,7 @@ int main(int argc, char** argv)
         }
         else
         {
-            // In a normal installation, we can infer the path
-            const sys::Path progDirname =
-                sys::Path::splitPath(progname).first;
-            const sys::Path schemaPath = progDirname.join("..").join("..").
-                join("conf").join("schema").join("six");
-            schemaPaths.push_back(sys::Path::absolutePath(schemaPath));
+            schemaPaths.push_back(six::findSchemaPath(progname));
         }
 
         std::auto_ptr<six::sicd::ComplexData> fileComplexData;
@@ -82,11 +78,8 @@ int main(int argc, char** argv)
         reader.load(stream, schemaPaths);
 
         streamComplexData = six::sicd::Utilities::getComplexData(reader);
-        six::sicd::Utilities::getWidebandData(reader, *(streamComplexData.get()), streamWidebandData);
-
-        // This tells the reader that it doesn't
-        // own an XMLControlRegistry
-        reader.setXMLControlRegistry(NULL);
+        six::sicd::Utilities::getWidebandData(reader, *streamComplexData,
+                streamWidebandData);
 
          if (*streamComplexData != *fileComplexData)
         {
--- a/six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
+++ b/six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
@@ -1,0 +1,120 @@
+/* =========================================================================
+* This file is part of six-c++
+* =========================================================================
+*
+* (C) Copyright 2004 - 2016, MDA Information Systems LLC
+*
+* six-c++ is free software; you can redistribute it and/or modify
+* it under the terms of the GNU Lesser General Public License as published by
+* the Free Software Foundation; either version 3 of the License, or
+* (at your option) any later version.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU Lesser General Public License for more details.
+*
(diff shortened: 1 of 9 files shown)


six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
```diff
@@ -0,0 +1,120 @@
...
+        six::NITFReadControl reader;
+        reader.setXMLControlRegistry(&xmlRegistry);
+        reader.load(stream, schemaPaths);
+
+        streamComplexData = six::sicd::Utilities::getComplexData(reader);
+        six::sicd::Utilities::getWidebandData(reader, *(streamComplexData.get()), streamWidebandData);
+
+        // This tells the reader that it doesn't
+        // own an XMLControlRegistry
+        reader.setXMLControlRegistry(NULL);
```
asylvest: Why is this needed?  I don't see the reader trying to delete it anywhere...

six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
```diff
@@ -0,0 +1,120 @@
...
+        xmlRegistry.addCreator(six::DataType::COMPLEX,
+            new six::XMLControlCreatorT<
+            six::sicd::ComplexXMLControl>());
+
+        six::NITFReadControl reader;
+        reader.setXMLControlRegistry(&xmlRegistry);
+        reader.load(stream, schemaPaths);
+
+        streamComplexData = six::sicd::Utilities::getComplexData(reader);
+        six::sicd::Utilities::getWidebandData(reader, *(streamComplexData.get()), streamWidebandData);
```
asylvest: Can just use `*streamComplexData`

six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+        std::vector<std::string> schemaPaths;
+        if (argc == 3)
+        {
+            schemaPaths.push_back(argv[2]);
+        }
+        else
+        {
+            // In a normal installation, we can infer the path
+            const sys::Path progDirname =
```
asylvest: Can you make a utility function for this somewhere common that we can use for programs like this?

six/modules/c++/six/include/six/NITFReadControl.h
```diff
@@ -222,6 +223,10 @@ class NITFReadControl : public ReadControl
 
         return (iCat == "LEG");
     }
+
+    // We need this for one of the load overloadings
+    // to prevent data from being deleted prematurely
+    mem::ScopedCopyablePtr<nitf::IOStreamReader> mStreamReader;
```
asylvest: So it's a tradeoff... this puts a weird special case as a member variable but in exchange doesn't change the public APIs for the `load(nitf::IOInterface& )` cases.  But what happens if we have another `CustomIO` or other overloading that also needs a smart pointer.  I think what would be cleaner is to have this simply hold a `mem::SharedPtr<nitf::IOInterface>` and always set it for all the overloadings.  That will change a couple load() overloadings, but I think it's worth it.  Thoughts?

JonathanMeans:
I’m a little concerned that having that for every IOInterface would be somewhat misleading and make it harder to make clear what exactly we’re guarding against. But on the other hand, if there’s just a smart pointer for that special case, then someone tries to load the same reader with something else, then the stream from the old load would still be open, which seems like it could be bad. So I agree that a shared pointer to IOInterface is the way to go.

JonathanMeans:
This is fixed.

six/modules/c++/six/source/Utilities.cpp
```diff
@@ -1101,6 +1101,34 @@ std::auto_ptr<Data> six::parseDataFromString(const XMLControlRegistry& xmlReg,
...
 }
 
+std::string six::findSchemaPath(const std::string& progname)
+{
+    sys::OS os;
+    std::string currentDir = os.getCurrentExecutable(progname);
+
+    // Arbitrary depth to prevent infinite loop in case
+    // of weird project structure
+    const size_t MAX_DEPTH = 5;
```
asylvest: This can be `static`

six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
```diff
@@ -0,0 +1,113 @@
+/* =========================================================================
+* This file is part of six-c++
```
asylvest: Should be `six.sicd-c++`

six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
```diff
@@ -0,0 +1,113 @@
...
+* =========================================================================
+*
+* (C) Copyright 2004 - 2016, MDA Information Systems LLC
+*
+* six-c++ is free software; you can redistribute it and/or modify
```
asylvest: Here too

six/modules/c++/six.sicd/tests/test_load_from_input_stream.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+        six::NITFReadControl reader;
+        reader.setXMLControlRegistry(&xmlRegistry);
+        reader.load(stream, schemaPaths);
+
+        streamComplexData = six::sicd::Utilities::getComplexData(reader);
+        six::sicd::Utilities::getWidebandData(reader, *streamComplexData,
+                streamWidebandData);
+
+         if (*streamComplexData != *fileComplexData)
```
asylvest: One extra space here

asylvest:
This looks good.  Let me know when those couple trivial things are fixed and I'll merge.

JonathanMeans:
They’re fixed.

PR #126, opened 2016-11-17
Add methods to convert DateTime to Python datetime object

JonathanMeans:


six/modules/python/six/source/six.i
```diff
@@ -100,6 +102,40 @@ def setValue(self, *args):
...
+                hour, minute, second, microsecond);
+        return nitfDateTime;
+    }
+
+    %pythoncode
+    %{
+        @staticmethod
+        def fromPythonDateTime(pyDatetime):
+            second = pyDatetime.second + (pyDatetime.microsecond / 1e6);
+            sixDatetime = DateTime()
```
asylvest: You can actually use the `DateTime` constructor that lets you send in all these things in upfront.  This will be (very slightly) more efficient since the default constructor gets the current date/time and there's no need to do that.

six/modules/python/six/tests/testDateTime.py
```diff
@@ -0,0 +1,70 @@
...
+def convertToPython():
+    original = DateTime()
+    converted = original.toPythonDateTime()
+    try:
+        assert original.getYear() == converted.year
+        assert original.getMonth() == converted.month
+        assert original.getDayOfMonth() == converted.day
+        assert original.getHour() == converted.hour
+        assert original.getMinute() == converted.minute
+        assert int(original.getSecond()) == converted.second
```
asylvest: You can test that the microseconds match in these two tests too right?

JonathanMeans:
These are fixed.

asylvest:
Looks good

PR #127, opened 2016-11-17
 Add makewheel and package commands to docbook 

JonathanMeans:


asylvest:
This looks good - thanks

PR #128, opened 2016-11-17
Add round-trip to make sure GeoTIFFReader/Writer works

JonathanMeans:


six/modules/c++/six.sidd/tests/test_geotiff.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include "six/Types.h"
+
+namespace
+{
+static const std::string OUTPUT_NAME("GeoTIFFTest");
+static const size_t DATA_LENGTH = 100;
+static const size_t DATA_SIZE_IN_BYTES = DATA_LENGTH * sizeof(sys::Int16_T) / sizeof(six::UByte);
+static const size_t BYTES_PER_PIXEL = sizeof(sys::Int16_T);
+
+void generateData(mem::ScopedArray<sys::Int16_T>& data)
```
asylvest: Since you're not doing anything with the fact that this is a smart pointer, have the function just take in a raw pointer

six/modules/c++/six.sidd/tests/test_geotiff.cpp
```diff
@@ -0,0 +1,168 @@
...
+{
+    for (size_t ii = 0; ii < DATA_LENGTH; ++ii)
+    {
+        data[ii] = sys::Int16_T(static_cast<sys::Int16_T>(ii));
```
asylvest: You're already doing a `static_cast`, so don't need the outer `sys::Int16_T()` here

JonathanMeans:
These are fixed.

asylvest:
Looks good

PR #129, opened 2016-11-20
Fixing a bug where OPT_BUFFER_SIZE was not actually used

asylvest:
Also increased default buffer size from 8 KB to 8 MB.

FYI @JonathanMeans


asylvest:
Not sure when or why this was originally commented out but it's been that way for at least a year.

PR #135, opened 2016-12-13
Fix CSM plugin parsing for NitfIsd types

JonathanMeans:


asylvest:
Ugh - this was a bug.  This broke when we updated for SICD 1.2 which fields we were keying off of.

I started with your updates here but simplified them (`NITFReadControl` can do more of the checks), tried to make really explicit in comments what we were doing each step since we can't use NITRO, and added some checks for cases where the DES is shorter than expected (could happen if someone is forming a `NitfIsd` themselves and fills it out wrong... like I did :o) ).

PR #137, opened 2016-12-18
Fix bug in CSM pixel treatment for SICDs

asylvest:
The CSM convention is 0,0 is the upper-left corner of the pixel whereas SICD convention is 0,0 is the center of the pixel.  This needs to be accounted for in the SICDSensorModel - prior to this fix, all SICD measurements were slightly off.  If you measure a SICD via CSM at ImageData\SCPPixel + 0.5 and run this through imageToGround(), you should now get a value extremely close to the SICD's GeoData\SCP\ECF.


asylvest:
@JonathanMeans

Issue #43, opened 2016-01-21
Create __init__.py files on install


JonathanMeans:
In order for the Python modules to work (at least on Windows), `__init__.py` files are needed in `install/lib/site-packages/coda` and `install/lib/site-packages/pysix`. Currently, these are not present after a vanilla `waf install` command.

chvink:
I'm kind of surprised but it actually worked for me. I just built it with W7, whatever compiler Visual Studio Express 2013 ships with, and python 2.7.9 from anaconda. Until this is figured out you can probably get away with making the files by hand.

mbojrab:
I needed to add this to my PYTHONPATH to get it to work, but also was surprised no __init__.py was required. I think its worth adding the __init__.py because its standard in module creation. We should also consider creating a setup.py installer (directly into the python library) or creating a mechanism for installing the egg from the geohub page using pip or something similar. Having a "pip install/update six-library" or "conda isntall/update six-library" would be very beneficial to us and our users. I'd use it.

chvink:
I'm getting an `__init__.py` when I built and installed - it's definitely intended to be there. 

I did look into methods of packaging and creating a setup.py a long time ago, but I haven't had the time to follow through.

mbojrab:
I just checked my recent six install and the `__init__.py` are populated as expected. <install>\Lib\site-packages\pysix and <install>\Lib\site-packages\coda. It has been some time since I last checked.

I perform one call for my build sequence:
`waf configure --prefix ..\six64.release -o target64.release --disable-warnings install`

JonathanMeans:
The __init__.py files are populating for me now, as well. If I did anything differently, I am not aware of it.

chvink:
@JonathanMeans

I just managed to reproduce this problem. I had an issue in a swig binding that caused the build to fail. After fixing that issue, subsequent `./waf install`s would not create the `__init__.py`. A distclean and reconfigure seems to put everything back in working order, but something is not working quite as intended so I'm reopening this issue.

Issue #119, opened 2016-10-24
Add utility to parse ComplexData from either SICD NITF or XML


msciaramitaro:
Add a utility method that will parse a ComplexData object from a given filename.  That file can be formatted as either a standard SICD NITF or a SICD XML text file (a single method to support both).

@asylvest @JonathanMeans

PR #143, opened 2016-12-29
let getComplexData read from XML

JonathanMeans:
https://github.com/ngageoint/six-library/issues/119

@msciaramitaro


six/modules/python/six.sicd/source/six_sicd.i
```diff
@@ -306,6 +308,19 @@ def readRegion(inputPathname, startRow, numRows, startCol, numCols, schemaPaths
     getWidebandRegion(inputPathname, schemaPaths, complexData, startRow, numRows, startCol, numCols, widebandBuffer)
 
     return widebandData, complexData
+
+def writeXML(complexData, pathname):
```
asylvest: Typically I'd just think of Swig code going in the .i file but I guess we don't have a good precedent for utility functions like this going elsewhere?  Would it be straightforward to put this in its own .py file somewhere that waf would still install?  If that's a big pain, it's ok to leave it here for now.

JonathanMeans:
I've made the changes. I think I fixed https://github.com/ngageoint/six-library/issues/43 as well.

six/modules/python/six.sicd/source/six_sicd.i
```diff
@@ -306,12 +308,13 @@ def readRegion(inputPathname, startRow, numRows, startCol, numCols, schemaPaths
...
 
     return widebandData, complexData
+
+
 def writeAsNITF(outFile, schemaPaths, complexData, image):
     writeNITF(outFile, schemaPaths, complexData,
         image.__array_interface__["data"][0])
 
 def readFromNITF(pathname, schemaPaths=VectorString()):
-    pathname = pathname + ".nitf"
```
asylvest: This seemed like a strange line to be in there... I assume that was in there by accident before and you removed it with intention here?

asylvest:
I see that `__init__.py` got touched but GitHub isn't showing any diff for it... did you change permissions on it or something or what actually changed?  How did this (or something else) affect #43?

six/modules/python/six.sicd/source/six_sicd.i
JonathanMeans: There were actually a number of read/write functions that append extensions, so I wouldn't say I put it there by accident. But this is an intentional removal, yes.

JonathanMeans:
All I did was create an empty __init__.py. Then waf should put it in the correct location when it installs. 

Somehow, the __init__.py gets created if you do `waf clean configure [...] install`, but if you just try to remove the install dir then do `waf install`, that's when the error would show up.

JonathanMeans:
The __init__.py is gone from this now.

asylvest:
This is good to go

PR #144, opened 2017-01-19
Fix error in SIDD CSM test

JonathanMeans:


asylvest:
Ahh well that's good news that that's all it was!

PR #146, opened 2017-01-25
Header size now calculated correctly

carlgbach:
Function CPHDWriter::writeMetadata() had a defect in which the header size was calculated prior to the addition of the Classification and Release Information strings. If either of these strings were added the header's offsets to the other sections of the CPHD file were incorrect, resulting in a garbled CPHD file.


asylvest:
Ugh - yep, I agree that was a bug.  Fix looks good.

FYI @JonathanMeans

Issue #147, opened 2017-01-29
Python CPHD lacks CPHDWriter


darkstar007:
Hi,

It appears that the python wrapper for CPHD does not have a CPHDWriter class in it - or have I missed something? This makes writing out CPHD files from python impossible?

Thanks,

Matt

asylvest:
Matt - that is correct... when we put in the CPHD bindings, at the time we only needed reading support.  Do you have a current need for writing support?  If so, let me know the urgency and timeframe and we'll try to prioritize it appropriately on our end.

darkstar007:
Hi,

I was wanting to write a simple simulator to test the processing chain of our new LF SAR system. So it would be really good if it could be done in the next 6 weeks or so.

Thanks,

Matt

asylvest:
Thanks for the timeframe Matt.  We'll see what we can do and will keep you posted...

asylvest:
@darkstar007 - see @JonathanMeans's updates in #153.  This has been merged into master... let us know if it meets your needs.

darkstar007:
OK, cool. I'll pull and check over the weekend.

darkstar007:
Hi.

I've spent some time on this in the last week or so, and have to say its been a hugely frustrating experience. This is mainly due to the lowlevel nature of the API, the lack of documentation and the poor error reporting.

I've yet to actually write out a CPHD file. My current issue is:

```
Traceback (most recent call last):
  File "./sim.py", line 255, in <module>
    writer.writeCPHD(sys.argv[1], vdata, coda_types.RowColSizeT(vdata.shape[0], vdata.shape[1]), vbm.toBuffer(0))
  File "/usr/local/lib/python2.7/dist-packages/pysix/cphd.py", line 2047, in write
    self.write(pathname)
  File "/usr/local/lib/python2.7/dist-packages/pysix/cphd.py", line 2009, in write
    return _cphd.CPHDWriter_write(self, *args)
RuntimeError: Attempted use of uninitialized double value
```

Which tells me I'm missing a variable in the metadata class, but no hints as to which one. So I now need to go through the spec and see what I've missed off.

Can I suggest changing the CPHDWriter.writeCPHD function to be something along the lines of:

```python
def write(self, pathname, vph_data, vbm, channel):
   """
    Function to write out video phase history data to a given filename.
        pathname - the filename (a string) to write the data to
        vph_data - a 2D numpy array (numpy.complex64 type) of size [npulse, nsamps]
[...]

PR #153, opened 2017-02-10
Add CPHDWriter to Python layer

JonathanMeans:
This allows for writing `std::complex<float>` CPHD files. The Python-layer writer and reader are currently restricted to this overload, because numpy doesn't have a datatype corresponding to `complex<int16>` or `complex<int8>.` I'm assuming since the reader's been fine how it is, there's no immediate need to support the other two overloads.  If there is, that might take a bit of thinking.

#147 
@darkstar007

### DIFF
--- a/six/modules/c++/cphd/source/CPHDWriter.cpp
+++ b/six/modules/c++/cphd/source/CPHDWriter.cpp
@@ -32,13 +32,19 @@ namespace cphd
 CPHDWriter::DataWriter::DataWriter(io::FileOutputStream& stream,
             size_t numThreads) :
     mStream(stream),
-    mNumThreads(numThreads)
+    mNumThreads(initNumThreads(numThreads))
 {
 }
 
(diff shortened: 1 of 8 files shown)


six/modules/c++/cphd/source/CPHDWriter.cpp
```diff
@@ -32,14 +32,20 @@ namespace cphd
...
-    mNumThreads(numThreads)
+    mNumThreads(initNumThreads(numThreads))
 {
 }
 
 CPHDWriter::DataWriter::~DataWriter()
 {
 }
 
+size_t CPHDWriter::initNumThreads(size_t numThreads)
```
asylvest: This function is simple enough and won't get reused anywhere that I think it's actually simpler just to do this right in the initializer list so the reader doesn't have to go find this function.

JonathanMeans:
Fixed.

PR #155, opened 2017-02-20
Fix error in writing blocked, segmented SIDDs

JonathanMeans:


asylvest:
Looks like you missed pushing NITFWriteControl.cpp and/or any other changes you made (including calling the new test program from the Python tests?)

JonathanMeans:
Oops! Sorry about that. Try it now.

six/modules/c++/six.sidd/tests/test_sidd_blocking.cpp
```diff
@@ -0,0 +1,194 @@
...
+            six::sidd::DerivedXMLControl>());
+
+        const std::string installPathname(argv[1]);
+        TempFileWithExtension multiImageSidd(".nitf");
+        TempFileWithExtension roundTrippedSidd(".nitf");
+        getMultiImageSIDD(installPathname, multiImageSidd.pathname());
+        roundTripAndBlock(installPathname,
+                multiImageSidd.pathname(), roundTrippedSidd.pathname());
+
+        if (checkBlocking(multiImageSidd.pathname(),
```
asylvest: This feels a little hokey to be calling other C++ applications inside C++ and then within Python calling this but I get why you did it this way and it does make the test overall simple... the alternative would maybe be to make function calls that do 98% of the implementation of round_trip_six and test_create_sidd_from_mem but it's probably not worth the trouble.

six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -866,11 +866,14 @@ void NITFWriteControl::save(
...
 
                 nitf::ImageSource iSource;
-                const size_t bandSize = numCols * info.getData()->getNumRows();
+                const NITFSegmentInfo segmentInfo = imageSegments[jj];
+                const size_t bandSize = numCols * segmentInfo.numRows;
 
                 for (size_t chan = 0; chan < numChannels; ++chan)
                 {
-                    nitf::MemorySource ms(imageData[i], bandSize,
+                    nitf::MemorySource ms(imageData[i] +
```
asylvest: The offset needs to also be multiplied by pixelSize.  Rather than doing the pointer arithmetic here, I *think* you can simply pass in `segmentInfo.firstRow * numCols` to the last constructor parameter here (that's set to 0 right now).  But take a look at the underlying code to verify that's true.

To catch this, can you add a test that writes out a multi-image segment SIDD that's 16 bits / pixel?

six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -866,11 +866,14 @@ void NITFWriteControl::save(
                 iWriter.setWriteCaching(1);
 
                 nitf::ImageSource iSource;
-                const size_t bandSize = numCols * info.getData()->getNumRows();
+                const NITFSegmentInfo segmentInfo = imageSegments[jj];
+                const size_t bandSize = numCols * segmentInfo.numRows;
```
asylvest: The NITRO comment is ambiguous here - make sure that what the `MemorySource` constructor wants is numPixels (which is what `bandSize` is) vs. number of bytes.

six/modules/c++/six.sidd/tests/test_sidd_blocking.cpp
asylvest: We'll want a 16 bits / pixel SIDD test in here too based on the MemorySource issue below.

six/modules/c++/six/source/NITFWriteControl.cpp
JonathanMeans: It looks like the last parameter has to do with the number of pixels in between columns or something like that. I think I could add the offset to the 4th parameter, but I think it reads a bit more clearly if the offset due to the segment is separate from the offset due to the channel. Everything else is fixed.

asylvest: OK no prob - just doing it via the pointer arithmetic is fine.  I'll take a look at the rest of the changes here shortly.

six/modules/c++/six.sidd/tests/test_sidd_blocking.cpp
```diff
@@ -0,0 +1,224 @@
...
+                roundTrippedMultiBandSidd.pathname());
+
+        if (checkBlocking(multiImageSidd.pathname(),
+                roundTrippedSidd.pathname()) and
+            checkBlocking(multiBandMultiImageSidd.pathname(),
+                roundTrippedMultiBandSidd.pathname()))
+        {
+            return 0;
+        }
+        return 1;
```
asylvest: Don't need this return 1 since you fall through below

six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -956,6 +961,7 @@ void NITFWriteControl::addDataAndWrite(
     int deWriterIndex = static_cast<int>(mContainer->getNumData());
     for (size_t ii = 0; ii < mSegmentWriters.size(); ++ii)
     {
+        // TODO: Investigate
```
asylvest: I don't think you need this todo - I *think* this section will work fine

asylvest:
Since these were trivial changes I just went ahead and made them so we could merge this in.  This all looks good to me.

PR #157, opened 2017-02-20
Convenience overload to make it easy to read entire image

JonathanMeans:


asylvest:
So, at first I was just going to add in a few more Doxygen comments (oh and I made it non-virtual, then realized it could just go in the base `ReadControl` class).

Then the more I looked at the new `interleaved()` overloading, I got concerned that we were adding in another function that the caller could leak memory if they weren't careful and I really thought it should have you pass in a `ScopedArray` to take ownership over that memory.

Then I looked at the update you'd made to the test program that used this new overloading and I saw that with the change, it was re-allocating memory now rather than reusing it.  It doesn't *really* matter for a test like this, but it was still bugging me... the old `read()` implementation in `test_byte_swap` can be still way-simplified since you can just default construct `Region`... all you need to do is set its buffer.

I still think there's value in having this new overloading but I know I re-jiggered a lot of stuff from what you originally put in... take a look and let me know what you think.

JonathanMeans:
I'm confused now. I realized last night that a default-constructed Region is set up to read the entire image...  so I'm not sure it makes sense to have a separate overloading so just that can be read safely. If anything what about 

`void interleaved(Region& region, size_t imageNumber, mem::ScopedArray<UByte>& buffer)`

and then comments making clear that this is the better option, and a default-constructed Region does the right thing?

asylvest:
Yeah I don't disagree since normally you're going to want to know how big this thing is that you just read in (otherwise without the `Region` you need to query `six::Data` yourself).  For consistency I'd still have it return `UByte*` I suppose.

Go ahead and make that update.

JonathanMeans:
The array was feeling a little awkward to use, so I tried a template.

Unrelated: I'm having trouble making SIX fail on Windows locally. May be stuck slowly debugging it through Jenkins.

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -118,22 +118,13 @@ void write(const sys::Int16_T* data, bool useStream, bool byteSwap)
...
 }
 
-void read(const std::string& filename, sys::Int16_T* data)
+void read(const std::string& filename, mem::ScopedArray<sys::Int16_T>& data)
 {
-    six::NITFReadControl reader;
-    reader.load(filename);
-    mem::SharedPtr<const six::Container> container = reader.getContainer();
-    const six::Data* const inData = container->getData(0);
+    std::auto_ptr<six::ReadControl> reader(new six::NITFReadControl());
```
asylvest: Why not use NITFReadControl on the stack like before?

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -118,22 +118,13 @@ void write(const sys::Int16_T* data, bool useStream, bool byteSwap)
...
-    const size_t numRows = inData->getNumRows();
-    const size_t numCols = inData->getNumCols();
     six::Region region;
-    region.setStartRow(0);
-    region.setStartCol(0);
-    region.setNumRows(numRows);
-    region.setNumCols(numCols);
-    region.setBuffer(reinterpret_cast<six::UByte*>(data));
-    reader.interleaved(region, 0);
+    reader->interleaved(region, 0, data);
```
asylvest: This will reallocate rather than reuse the memory...

six/modules/c++/six/include/six/ReadControl.h
```diff
@@ -108,12 +109,43 @@ class ReadControl
...
+     * This will be allocated by this function.
+     *
+     * \return Buffer of image data.  This is simply equal to buffer.get() and
+     * is provided as a convenience.
+     */
+    template<typename T>
+    UByte* interleaved(Region& region, size_t imageNumber,
+            mem::ScopedArray<T>& buffer)
+    {
+        buffer.reset(reinterpret_cast<T*>(interleaved(region, imageNumber)));
```
asylvest: So... if they pass in a region with a non-NULL buffer, they will probably think they still own the memory and double delete it.  I wish the region held a smart pointer but it doesn't right now.  Any thoughs?

asylvest: I like this templated - might as well make the return type T* for convenience too

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
JonathanMeans: It wasn't seeing it in the base class.

asylvest: Oh yeah... add a `using ReadControl::interleaved` in NITFReadControl... do some Googling to see why that's needed.

six/modules/c++/six/include/six/ReadControl.h
JonathanMeans: I am not overfond of the `Region` holding any sort of pointer at all, as it makes it tricky to use the same `Region` for multiple reads.  If we didn't have to worry about breaking anything, I'd suggest removing it altogether. 

I'm trying to think through passing a Region that already has a buffer. I would hope that  the smart pointer parameter and the documentation would make it clear that that's what owns the memory at the end. Presumably the caller would already have it stored as a ScopedArray and not just a smart pointer, so that's what they'd end up passing as the third argument, and nothing would change.

Or if we change Region to hold a smart pointer, we don't even need the new overload, and there's one less pointer to the same data getting passed around.

JonathanMeans: I am not overfond of the `Region` holding any sort of pointer at all, as it makes it tricky to use the same `Region` for multiple reads.  If we didn't have to worry about breaking anything, I'd suggest removing it altogether. 

I'm trying to think through passing a Region that already has a buffer. I would hope that  the smart pointer parameter and the documentation would make it clear that that's what owns the memory at the end. Presumably the caller would already have it stored as a ScopedArray and not just a smart pointer, so that's what they'd end up passing as the third argument, and nothing would change.

Or if we change Region to hold a smart pointer, we don't even need the new overload, and there's one less pointer to the same data getting passed around.

six/modules/c++/six/include/six/ReadControl.h
```diff
@@ -108,12 +109,43 @@ class ReadControl
...
      *  Once read, the image buffer is set in both the region pointer,
      *  and in the return value, for convenience
+     *
+     *  For safety, prefer the overload below.
      */
     virtual UByte* interleaved(Region& region, size_t imageNumber) = 0;
 
     /*!
+     *  This function reads in the image area specified by the region.
+     *  If you want us to use a work buffer, you can set it through region.
```
asylvest: Since `buffer` takes ownership of this, spell out in the comments here what happens if you do in fact set that work buffer through region (i.e. you don't own the memory after this call - the ScopedArray does)

six/modules/c++/six.sidd/tests/test_geotiff.cpp
```diff
@@ -102,33 +102,25 @@ void write(const sys::Int16_T* data)
     writer.save(reinterpret_cast<const six::UByte*>(data), OUTPUT_NAME);
 }
 
-void read(const std::string& filename, sys::Int16_T* data)
+void read(const std::string& filename, mem::ScopedArray<sys::Int16_T>& data)
 {
-    six::sidd::GeoTIFFReadControl reader;
+    six::NITFReadControl reader;
```
asylvest: Does this test program really get run?  This can't be right... it should still be a `GeoTIFFReadControl` right??

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
```diff
@@ -118,22 +118,13 @@ void write(const sys::Int16_T* data, bool useStream, bool byteSwap)
...
-    const size_t numRows = inData->getNumRows();
-    const size_t numCols = inData->getNumCols();
     six::Region region;
-    region.setStartRow(0);
-    region.setStartCol(0);
-    region.setNumRows(numRows);
-    region.setNumCols(numCols);
-    region.setBuffer(reinterpret_cast<six::UByte*>(data));
-    reader.interleaved(region, 0);
+    reader.interleaved(region, 0, data);
```
asylvest: I'm still a little hung up on that we're not re-using the memory now... to me, just adding one more line (the `region.setBuffer()` line from before) and passing in a raw pointer here rather than the `ScopedArray` really is pretty legible code, but it is the case that we're talking a trivial amount of memory and this is a test program, so if you think this does improve readability, I'm good with it.

six/modules/c++/six/include/six/ReadControl.h
```diff
@@ -108,12 +109,43 @@ class ReadControl
...
+     *
+     *  To simply read the entire image, pass a default-constructed
```
asylvest: Also add a comment somewhere in here that there's no casting that happens... for example, if it's an integer SICD on disk and you pass in a `ScopedArray<std::complex<float>>`, we're not going to cast the data to complex float for you; it'll just do the wrong thing.

six/modules/c++/six.sidd/tests/test_geotiff.cpp
```diff
@@ -102,33 +102,25 @@ void write(const sys::Int16_T* data)
...
+    reader.interleaved(region, 0, data);
 }
 
 bool run()
 {
     mem::ScopedArray<sys::Int16_T> imageData(new sys::Int16_T[DATA_LENGTH]);
     generateData(imageData.get());
 
-    mem::ScopedArray<sys::Int16_T> testData(new sys::Int16_T[DATA_LENGTH]);
+    //mem::ScopedArray<sys::Int16_T> testData(new sys::Int16_T[DATA_LENGTH]);
```
asylvest: Don't need this line anymore

six/modules/c++/six.sidd/tests/test_byte_swap.cpp
JonathanMeans: Sorry, I'd lost track of this one! I've pushed up the fixes.

asylvest:
Alright cool - this all looks good

PR #159, opened 2017-02-21
Remove pairs to fix Solairs error

JonathanMeans:


asylvest:
Is it just that you need to include `<utility>` to get `std::pair`?  I don't see that included in this file...

JonathanMeans:
No, it doesn't like ScopedArray not having a copy constructor.

asylvest:
Gotcha - Solaris is lame.

PR #162, opened 2017-02-23
Improve error messages in ProjectionModel ctors

JonathanMeans:


six/modules/c++/scene/source/ProjectionModel.cpp
```diff
@@ -62,11 +62,19 @@ ProjectionModel(const Vector3& slantPlaneNormal,
     mSlantPlaneNormal(slantPlaneNormal),
     mSCP(scp),
     mARPPoly(arpPoly),
-    mARPVelPoly(mARPPoly.derivative()),
     mTimeCOAPoly(timeCOAPoly),
     mLookDir(lookDir),
     mErrors(errors)
 {
+    try
```
asylvest: This will throw when?  When the poly is empty so order() throws?  If we really need to guard against that - I take it you had some confusing exception message you tracked down to this - can you make it one standalone function in the unnamed namespace that does this?  This way you can reuse the code and still keep things in the initializer list.  The function would take in a string for the poly name.

six/modules/c++/six.sicd/source/GeoData.cpp
```diff
@@ -114,20 +114,30 @@ bool GeoData::validate(logging::Logger& log) const
     bool valid = true;
     std::ostringstream messageBuilder;
 
-    // 2.10
-    scene::LLAToECEFTransform transformer;
-    Vector3 derivedEcf = transformer.transform(scp.llh);
-    double ecfDiff = (scp.ecf - derivedEcf).norm();
-
-    if (ecfDiff > ECF_THRESHOLD)
+    if (Init::isUndefined(scp.llh) ||
```
asylvest: The pull request title is for the ProjectionModel constructor... are these intentional?

six/modules/c++/scene/source/ProjectionModel.cpp
JonathanMeans: Yes, it threw when the poly was empty, which took a bit of time to track down.  I've made the change.

six/modules/c++/six.sicd/source/GeoData.cpp
JonathanMeans: Yes. I ran into a couple segfaults after I made the request and thought it would fit.

asylvest:
Looks good

PR #166, opened 2017-03-06
Add utility for generate fake ComplexData

JonathanMeans:
I've run into several testing situations where I want to check that the right thing is happening with part of a `ComplexData` object, but something else I don't care about isn't initialized and I get errors. This method returns a partially-filled ComplexData so that the only thing I need to fill out in my tests is the part I actually care about, and nothing else will complain. I grabbed data from a sample SICD because some of the things in `scene` can throw if the fields don't make sense with each other.

### DIFF
--- a/six/modules/c++/six.sicd/include/six/sicd/Utilities.h
+++ b/six/modules/c++/six.sicd/include/six/sicd/Utilities.h
@@ -329,6 +329,14 @@ public:
             const std::vector<std::string>& schemaPaths =
                     std::vector<std::string>(),
             logging::Logger* logger = NULL);
+
+    /*!
+     * Create a fake SICD that's populated enough for
+     * general testing code to run without throwing exceptions
+     *
+     * \return mock ComplexData object
+     */
+    static ComplexData createFakeComplexData();
 };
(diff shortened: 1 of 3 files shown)


six/modules/c++/six.sicd/include/six/sicd/Utilities.h
```diff
@@ -329,6 +329,14 @@ class Utilities
...
                     std::vector<std::string>(),
             logging::Logger* logger = NULL);
+
+    /*!
+     * Create a fake SICD that's populated enough for
+     * general testing code to run without throwing exceptions
+     *
+     * \return mock ComplexData object
+     */
+    static ComplexData createFakeComplexData();
```
asylvest: This seems like a good idea.  The one thing is can you change it to return an `auto_ptr` to a `ComplexData` to avoid the copy?

JonathanMeans: Done. I was thinking all our compilers would do NRVO here so the copy wouldn't actually happen, but it looks like Solaris doesn't....

PR #172, opened 2017-04-06
Add regression test for reads with large offset

JonathanMeans:


six/modules/c++/samples/test_large_offset.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+}
+
+int main(int argc, char** argv)
+{
+    try
+    {
+        if (argc != 2)
+        {
+            std::cerr << "Usage: " << argv[0] << " <Input NITF>\n";
```
asylvest: Rather than making this a test with an existing NITF, can't you use the fake data creation function (I guess you'll need to add one for SIDD) so that we can include this in the automated test suite?

six/modules/c++/samples/test_large_offset.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+#include <six/Container.h>
+#include <six/NITFReadControl.h>
+#include <six/NITFWriteControl.h>
+#include <six/sicd/ComplexXMLControl.h>
+#include <six/sidd/DerivedXMLControl.h>
+
+namespace
+{
+const size_t NUM_ROWS_IN_FIRST_SEGMENT = 47673; // By manual inspection
```
asylvest: These can all be `static` too

six/modules/c++/samples/test_large_offset.cpp
```diff
@@ -0,0 +1,150 @@
...
+    const std::vector<std::string> schemaPaths;
+    reader.load(pathname, schemaPaths);
+    std::auto_ptr<six::Data> data(reader.getContainer()->getData(0)->clone());
+
+    six::Region region;
+    region.setStartRow(ROWS_TO_SKIP);
+    region.setNumRows(data->getNumRows() - ROWS_TO_SKIP);
+    region.setStartCol(0);
+    reader.interleaved(region, 0);
+    return static_cast<sys::Int8_T>(region.getBuffer()[0]) == 1;
```
asylvest: So this test would have caught the old bug (which is good), but it's not as comprehensive as it could be.  Rather than just some memsets upfront, I think it would be a better test to fill in each pixel with incremental data (i.e. have a for loop that sets each pixel to the value of the for loop counter or something) and then here verify that the region that's read out matches what's expected for all pixel values rather than the first one.

six/modules/c++/samples/test_large_offset.cpp
asylvest: Oh I see that you did add this to the test suite... so that's good, but I think it would still be better to have it be totally self-contained and not require an input file.

JonathanMeans:
Ready.

six/modules/c++/samples/test_large_offset.cpp
```diff
@@ -0,0 +1,164 @@
...
+    return false;
+}
+}
+
+int main(int argc, char** argv)
+{
+    try
+    {
+        bool testPassed = runTest(six::DataType::COMPLEX);
+        testPassed = runTest(six::DataType::DERIVED);
```
asylvest: This logic needs to be a little different... right now if SICD failed but SIDD succeeded, we'd return 0

six/modules/c++/samples/test_large_offset.cpp
```diff
@@ -0,0 +1,164 @@
...
+    container->addData(data.release());
+
+    six::NITFWriteControl writer;
+    writer.setXMLControlRegistry(&registry);
+    writer.initialize(container);
+
+    mem::ScopedArray<six::UByte> imageData(new six::UByte[imageSize]);
+    for (size_t ii = 0; ii < imageSize; ++ii)
+    {
+        imageData[ii] = static_cast<six::UByte>(ii);
```
asylvest: This is definitely better, but we could still get unlucky and not catch a bug if our underlying reading logic was goofed up.  Since this is being treated as ubyte data, there are effectively only 256 unique values.  We could be reading from the totally wrong spot in the file and just happen to land on the same repeating values (granted, you only have a 1 / 256 chance of that).

To make really sure, how about changing the SICD check over to use `RE32F_IM32F` and then in this loop use incrementing floating point values (I assume you'll then need to cut the number of rows you're using in half to get the segment boundary math to work out)?  That way you'll be unique all the way through the image (might as well set the real and imaginary components to unique values too just to be sure... maybe one positive and one negative or something).  If you treat the SIDD as 16-bit values, that'll give you 2^16 rather than 2^8 unique values which isn't perfect but is good enough since we'll have an exhaustive SICD test.  Obviously you'll need a cast and different check down at the `checkNITF()` code that's in sync with this too.

JonathanMeans:
Good to go again.

six/modules/c++/samples/test_large_offset.cpp
```diff
@@ -0,0 +1,215 @@
...
+                return false;
+            }
+        }
+    }
+    return true;
+}
+
+bool runTest(const six::DataType& datatype)
+{
+    std::cerr << "Running offset test for datatype " << datatype.toString()
```
asylvest: Last nit.  This should be cout rather than cerr

JonathanMeans: Done.

asylvest:
Thanks - looks good

PR #174, opened 2017-04-18
Catch bad_alloc exception

JonathanMeans:


asylvest:
Ahh yeah, that makes sense

PR #179, opened 2017-04-21
Add overload to getOutputPlaneOffsetAndExtent

JonathanMeans:
There was a hiccough in merging in the derivedOutputPlane stuff. This will make more sense once that PR is in.

### DIFF
--- a/six/modules/c++/six.sicd/include/six/sicd/ComplexData.h
+++ b/six/modules/c++/six.sicd/include/six/sicd/ComplexData.h
@@ -323,14 +323,14 @@ public:
 
     /*!
      * Overload that allows caller to pass in their own AreaPlane
+     * \param areaPlane areaPlane to use for data
      * \param offset Output plane offset
      * \param extent Output plane extent
-     * \param areaPlane areaPlane to use for data
      */
     void getOutputPlaneOffsetAndExtent(
+            const AreaPlane& areaPlane,
             types::RowCol<size_t>& offset,
-            types::RowCol<size_t>& extent,
-            const AreaPlane& areaPlane) const;
+            types::RowCol<size_t>& extent) const;
 
     /*
      * Convert the slant plane pixel location into meters from the SCP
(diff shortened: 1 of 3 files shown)


six/modules/c++/six.sicd/include/six/sicd/ComplexData.h
```diff
@@ -314,11 +314,24 @@ class ComplexData: public Data
...
             types::RowCol<size_t>& extent) const;
 
+    /*!
+     * Overload that allows caller to pass in their own AreaPlane
+     * \param offset Output plane offset
+     * \param extent Output plane extent
+     * \param areaPlane areaPlane to use for data
+     */
+    void getOutputPlaneOffsetAndExtent(
+            types::RowCol<size_t>& offset,
```
asylvest: Output params should go last so areaPlane should be the first param

six/modules/c++/six.sicd/source/ComplexData.cpp
```diff
@@ -56,38 +56,42 @@ void ComplexData::getOutputPlaneOffsetAndExtent(
...
     extent.row = areaPlane.xDirection->elements;
     extent.col = areaPlane.yDirection->elements;
 
     if (imageFormation.get())
     {
         const std::string& segmentID =
                 imageFormation->segmentIdentifier;
+        try
+        {
+            const six::sicd::Segment& segment = areaPlane.getSegment(segmentID);
```
asylvest: Rather than a try/catch how about adding a hasSegment () method?  But also, when would this happen and you don't want to error out?  When segmentID is empty?  In that case, would it be better to explicitly check for that?

JonathanMeans: I'm lost on how the logic should be here. So if segmentID is empty, nothing should happen. If there is a segmentID, but it doesn't match anything in the list, it should throw. Is that correct?

I'm concerned that with an extra `hasSegment()` method, then to get a segment you'd have to loop over the list once to see if it's there, then again to actually get it. I'm not sure how long they normally are. What about replacing `getSegment()` with `getSegmentIndex()`?

asylvest: Yes, the logic you have above is right.

I think the logic you want here is to just replace the try/catch with a check to see if `segmentID` is empty.

I think we can avoid adding any new `hasSegment()` or `getSegmentIndex()` methods for now.  As a heads up, there will usually be a very small number of segments.   If it was 100, that would be insanely big.  So performance won't be an issue.  What I like about `getSegment()` is that the caller can be lazy with no error checking and act like it'll succeed (and in the rare case something is awry with the SICD or their logic, it will throw).

JonathanMeans:
Ready.

asylvest:
Looks good

PR #184, opened 2017-05-16
Add sample program for deriving output plane

JonathanMeans:


six/modules/c++/samples/get_output_plane_image.cpp
```diff
@@ -0,0 +1,159 @@
...
+        mem::ScopedArray<float> outputArray(new float[
+                plane.xDirection->elements * plane.yDirection->elements]);
+        // Iterate over output plane, grabbing the appropriate input point
+        for (size_t yy = 0; yy < plane.yDirection->elements; ++yy)
+        {
+            const six::Poly1D rowPoly = toSlantRow.atY(yy);
+            const six::Poly1D colPoly = toSlantCol.atY(yy);
+            for (size_t xx = 0; xx < plane.xDirection->elements; ++xx)
+            {
+                const size_t line = math::round(rowPoly(xx));
```
asylvest: Similarly I'd say change `line` to `inRow` and `sample` to `inCol` and then `inputPoint` to `inIdx` and `outputPoint` to `outIdx`

asylvest: Hmm, so `rowPoly` and/or `colPoly` could return a negative number.  Granted that'll actually get handled here because you'll cast to a `size_t` which'll set it to a bazillion but I think it would be more explicit - particularly since the point is to help people understand how to use these polys` - if you check for that

six/modules/c++/samples/wscript
```diff
@@ -7,6 +7,7 @@ def build(bld):
     samples = {'check_valid_six'           : 'cli six.sicd six.sidd',
                'crop_sicd'                 : 'cli six.sicd',
                'crop_sidd'                 : 'cli six.sidd',
+               'get_output_plane_image'    : 'cli io six six.sicd sio.lite',
```
asylvest: I'd rename this file to `project_slant_to_output.cpp`

JonathanMeans:
This is ready.

six/modules/c++/samples/project_slant_to_output.cpp
```diff
@@ -0,0 +1,168 @@
...
+ *
+ */
+
+/*
+ * This program gets a rough output plane from the SICD, and writes it as
+ * an SIO. No filtering is done. As a result, the output plane contains
+ * some massive values that skew the image. Therefore, when viewing the
+ * image, you will have to mitigate this. When viewing in MATLAB,
+ * for example, this may be done by
+ * imagesc(read_sio('outputPlane.sio', [0, mean(img(:))]));
```
asylvest: Need to show the assignment of img here

six/modules/c++/samples/project_slant_to_output.cpp
```diff
@@ -0,0 +1,168 @@
...
+        parser.addArgument("output", "Output SIO Pathname", cli::STORE,
+                           "output", "OUTPUT", 1, 1);
+
+        const std::auto_ptr<cli::Results> options(parser.parse(argc, argv));
+
+        const std::string sicdPathname(options->get<std::string>("input"));
+        const std::string outputPathname(options->get<std::string>("output"));
+        const size_t polyOrderX(options->get<size_t>("polyOrderX"));
+        const size_t polyOrderY(options->get<size_t>("polyOrderY"));
+        const std::vector<std::string> schemaPaths;
```
asylvest: Shouldn't we have an option for them to send in schemas?

JonathanMeans:
I've made the fixes.

asylvest:
Looks good

PR #185, opened 2017-05-18
Fixed logic in NITFImageInfo to account for image's possible larger s…

carlgbach:
…ize when blocked


asylvest:
I fixed some compilation and style issues but the logic all looked good to me.

PR #186, opened 2017-05-22
Handle the case where user requests more rows per block than allowed …

JonathanMeans:
…in a segment

(It feels odd for the `NITFImageInfo` ctor to take both `maxRows` and `maxSize` parameters. Doesn't one determine the other?)


asylvest:
Thanks for tracking this through.  I needed to fix the logic a little differently; did this in #187 so closing this one.

PR #188, opened 2017-05-22
Fix sizes passed into blocking

JonathanMeans:


JonathanMeans:
I've followed the logic through and convinced myself that the blocking is happening as desired. Is there any more concrete way to check this? When I pass in an image larger than a block, the numbers are showing that it needs an extra block to fit it, but I believe the actual blocking is handled by NITRO.

asylvest:
OK then you're probably good... in the final NITF, you want to look at numBlocksPerRow and numBlocksPerCol (you could dump it out via show_nitf) and confirm those are > 1.  Once you've done that, let me know and I'll merge.

JonathanMeans:
It checks out.

asylvest:
Cool - thanks.

PR #195, opened 2017-06-27
Fix rounding in test

JonathanMeans:
I ran this 10,000 times with no errors, so I am confident it is correct now.


six/modules/python/six/tests/testDateTime.py
```diff
@@ -58,7 +58,11 @@ def convertFromPython():
         assert(converted.getMinute() == original.minute)
         assert(int(converted.getSecond()) == original.second)
         microsecond = (converted.getSecond() - int(converted.getSecond())) * 1e6
-        assert(int(microsecond) == original.microsecond)
+        print('Original.microsecond:   {0}'.format(original.microsecond))
```
asylvest: Left some print statements in...

JonathanMeans:
Prints removed.

PR #198, opened 2017-07-27
 Catch XML errors for 0 channels in a more friendly manner  

JonathanMeans:


asylvest:
@JonathanMeans did this PR become OBE due to deciding we'd catch this in schema validation?

JonathanMeans:
Yes.

PR #200, opened 2017-09-04
Add ability for SIX to provide raw headers and offsets

asylvest:
Added a new method to `NITFWriteControl` that supports providing back the various NITF headers as raw bytes along with their associated file offsets.  This logic was largely reshuffled from `SICDWriteControl`; simplified `SICDWriteControl` to use this.

The advantage of this approach is it allows callers to have SIX figure out all the NITF file / header stuff and provide that info to them while they now have more control over how that SICD gets written out (there are some more complicated SICD writing scenarios we want to support where multiple machines worked on a portion of the SICD and are all writing their share of it).

Regenerated the Python bindings with Swig 3.0.12 (latest version)...


asylvest:
Yeah I started down that path... there is another branch where I was updating the various headers so you could ask them how many bytes they were.  But once I realized you'd want the headers as a raw byte stream and so at that point you knew the size anyhow, I figured it'd be one more thing to maintain.  That logic is in a branch if we ever want it though.

PR #204, opened 2017-10-05
Fix bug for gcc 4.4.7

JonathanMeans:
This is a workaround for this issue: https://stackoverflow.com/questions/5720359/no-matching-function-call-to-anonymous-enum


asylvest:
Fair enough

PR #206, opened 2017-10-13
Add pixel geolocation utility

JonathanMeans:


JonathanMeans:
Hang on, missed something tricky in the offset part.

JonathanMeans:
Good to go

six/modules/c++/samples/geolocate_pixel.cpp
```diff
@@ -0,0 +1,120 @@
...
+        const bool shadowsDown(options->get<bool>("shadowsDown"));
+        if (!six::sicd::AreaPlaneUtility::hasAreaPlane(*complexData))
+        {
+            six::sicd::AreaPlaneUtility::setAreaPlane(*complexData);
+        }
+
+        six::sicd::AreaPlane& plane = *complexData->radarCollection->area->plane;
+        if (shadowsDown)
+        {
+            plane.rotateCCW();
```
asylvest: You can't just always rotate CCW if you want shadows down... you need to check the orientation to see if you should rotate.  Probably worth adding a `rotateToShadowsDown()` method to the plane that does this check for you.  Note that if the orientation is `ARBITRARY`, you won't want to rotate it at all

JonathanMeans:
Made the changes.

six/modules/c++/samples/sicd_output_plane_pixel_to_lat_lon.cpp
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/*
+ * Provide a pixel coordinate and a SICD. This program will output the
+ * latitude, longitude, and latitude of the pixel in the output plane.
```
asylvest: second 'latitude' should be 'altitude'

six/modules/c++/samples/sicd_output_plane_pixel_to_lat_lon.cpp
```diff
@@ -0,0 +1,124 @@
...
+        if (!sys::OS().exists(sicdPathname))
+        {
+            std::cerr << sicdPathname << " does not exist.\n";
+            return 1;
+        }
+
+        std::vector<std::string> schemaPaths;
+        getSchemaPaths(*options, "--schemaPath", "schema", schemaPaths);
+
+        const double col = options->get<double>("col");
```
asylvest: Leftover `col`

JonathanMeans: Could this line have been from an outdated version? I don't see it...

asylvest: OK yeah I think the GitHub UI confused me or something - I agree that it's gone.

asylvest:
Looks good!

PR #211, opened 2017-11-04
Move image segment limits logic into NITRO

asylvest:
95% of what `NITFImageInfo` does with regards to breaking up a large image into multiple NITF image segments is more general than SICD/SIDD (the only aspect that is tied to it is the image segment corner calculation).

Abstracting this logic a bit and moving it into NITRO.  The logic in `ImageSegmentComputer` is a straight copy from SIX, with a bit of style cleanup and an attempt at more clear variable names in some spots.

Rather than putting out a separate NITRO pull request, merging that, then putting this one in, I thought it would be more clear to do this as one PR so you could see the SIX change at the same time.  Once this looks good, I'll put the NITRO piece out as a NITRO PR and then merge that into SIX like normal.

FYI @carlgbach


six/modules/c++/six/include/six/NITFImageInfo.h
```diff
@@ -150,7 +151,15 @@ class NITFImageInfo
         return mImageSegments;
     }
 
-    void addSegment(NITFSegmentInfo info)
+    /*!
+     * Adds a segment.  Note that the segmentation layout computed at
+     * construction time will no longer hold, so this should only be used
+     * when 'computeSegments' was false at construction or the segmentation
+     * layout is otherwise not being
```
JonathanMeans: not being what?

asylvest: Whoops - finished the sentence. :o)

asylvest:
Merged the NITRO portion of this in https://github.com/mdaus/nitro/pull/52

PR #212, opened 2017-11-14
Other programs won't install schemas to Python dir unless they ask fo…

JonathanMeans:
…r it.


asylvest:
Oh I see - I hadn't realized we were plopping it onto the six.sicd library via targets_to_add.  This fix looks good assuming that when we make the wheel we still get these installed.  Can you verify that if you haven't already?

JonathanMeans:
Confirmed.

asylvest:
Cool - thanks.

PR #217, opened 2017-12-19
Add option to pass NITFWriteControl to ctor

JonathanMeans:
Need this so we can control what goes in the NITF header when writing with SICDByteProvider.


six/modules/c++/six.sicd/source/SICDByteProvider.cpp
```diff
@@ -130,6 +130,28 @@ const void* SICDByteProvider::NITFBufferList::getBlock(
     return NULL;
 }
 
+size_t SICDByteProvider::getNumBytesPerRow(
+        const NITFWriteControl& writer) const
```
asylvest: This can be `static` rather than `const` right?

JonathanMeans: Done.

asylvest:
Looks good

PR #219, opened 2018-01-21
SIDD byte provider

asylvest:
* Moved the buffer list concept out of SICDByteProvider and into NITRO.
* Generalized SICDByteProvider into a base ByteProvider class with SICD and SIDD versions inheriting from it
* The main complexity introduced by SIDD is blocking, in particular with partial blocks.  Required some updates to NITRO's ImageBlocker because NITFWriteControl won't ever set numRowsPerBlock in a segment > numRows in that segment (which makes sense - there's no point), but this means we need to account for a potentially different numRowsPerBlock in each segment

Thought it would be easier to review the NITRO and SIX changes together.  Once this looks good, I'll perform a NITRO followed by a SIX pull request and merge.


six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -1289,12 +1298,15 @@ void NITFWriteControl::getFileLayout(
...
                      io::Seekable::CURRENT);
 
-    // Data extension segment
-    writer.writeInt64Field(desSubheaderLen, NITF_LDSH_SZ, '0',
-                         NITF_WRITER_FILL_LEFT);
+    // Data extension segments
+    for (size_t ii = 0; ii < numDESs; ++ii)
+    {
+        writer.writeInt64Field(desSubheaderLengths[ii], NITF_LDSH_SZ, '0',
+                             NITF_WRITER_FILL_LEFT);
```
JonathanMeans: Needs a couple extra spaces

externals/nitro/modules/c++/nitf/source/ImageBlocker.cpp
```diff
@@ -213,33 +221,38 @@ size_t ImageBlocker::getNumBytesRequired(size_t startRow,
     size_t numRowBlocks;
```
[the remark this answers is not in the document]
asylvest: Yeah originally they really were blocks but then I needed to change it to rows and couldn't come up with a better name.  I agree - updated to `totalNumRows`.

six/modules/c++/six/source/NITFWriteControl.cpp
asylvest: Fixed

asylvest:
NITRO functionality merged in https://github.com/mdaus/nitro/pull/74

PR #220, opened 2018-02-04
Generalize byte provider

asylvest:
This is the last generalization needed for the ByteProvider so it can be moved into NITRO (so we can use it with just TRE products... not just SICDs and SIDDs).  The TRE logic itself actually works out of the box as the file header and image subheaders already know how to write out the TRE data.

Essentially the logic from `six::ByteProvider` and `six::NITFWriteControl` moved to `nitf::ByteProvider`.  Added some checks to make sure graphics, labels, texts, and reserved extension segments aren't in the `nitf::Record`.

Review it here in one shot and then once all looks good I'll merge it in two steps.


asylvest:
Merged in NITRO https://github.com/mdaus/nitro/pull/76

PR #221, opened 2018-03-03
Fixing --disable-csm

asylvest:
`--disable-csm` should turn off building `six-csm`

### DIFF
--- a/six/projects/csm/wscript
+++ b/six/projects/csm/wscript
@@ -5,7 +5,7 @@ PLUGIN             = 'CSM'
 REMOVEPLUGINPREFIX = True
 
 import sys, os, re
-from waflib import Errors, Options
+from waflib import Errors, Options, Logs
 
 DIRS = 'external'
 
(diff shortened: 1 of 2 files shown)


six/projects/csm/wscript
```diff
@@ -20,42 +20,46 @@ def configure(conf):
...
+            else:
+                # If we are building with gcc on Windows, some of the CSM stuff
+                # doesn't compile.  No one is going to ever need to build CSM
+                # plugins in this paradigm as MSP and GXP won't be able to use
+                # them anyhow
+                enableCSM = False
+        elif re.match(osxRegex,sys.platform):
+            # The CSM build doesn't work on Mac... again, no one will need to build
+            # plugins there though
+            platformName='darwin'
```
JonathanMeans: Should there be some sort of warning to the config log in this case?

asylvest:
Added a warning.  Also moved the check so that you don't have to do `--disable-csm` on unsupported platforms... it'll just turn it off.

PR #227, opened 2018-04-17
Remove print statement

JonathanMeans:
Looks like the externals tests don't install by default. Should the SIX test suite be running these?


asylvest:
Nope that's with intention - only the bare minimum needed in the `externals` dir are built, so tests are skipped.

PR #231, opened 2018-05-14
Update siddByteProvider test to run multiple blocks in single write

JonathanMeans:


six/modules/c++/six.sidd/tests/test_sidd_byte_provider.cpp
```diff
@@ -511,57 +550,70 @@ void Tester<DataTypeT>::testMultipleWritesBlocked()
...
-            }
+        writeData.numBlocks = blocksThisWrite;
+        writeData.startRow = startRow;
+        writeData.numRows = numRows;
+        data.push_back(writeData);
+
+        startRow += numRows;
+    }
+
+    for (int ii = data.size() - 1; ii >= 0; --ii)
```
asylvest: Going to get a warning about signed/unsigned comparison here.  So instead, do something like...

```
for (size_t ctr = 0, ii = data.size() - 1; ctr < data.size(); ++ctr, --ii)
```

PR #238, opened 2018-06-12
Set J2K compression params with CompressedByteProvider

JonathanMeans:
Before I write the Doxygen, I'm not really sure what the new compressionLevel param is setting. In NITFWriteControl.cpp:307 it's transformed to `int comratInt = (int)((j2kCompression * nbpp * 10.0) + 0.5);` It looks like the passed value is meant to be compression percentage. If the comrat field is meant to map to openjpeg's compression factor, it would be nice if there were some way to set it to 1 (which means lossless compression for openjpeg). If it's not mirroring the compression factor, then I'm not really sure how to populate it.

### DIFF
--- a/six/modules/c++/six.sidd/include/six/sidd/CompressedSIDDByteProvider.h
+++ b/six/modules/c++/six.sidd/include/six/sidd/CompressedSIDDByteProvider.h
@@ -57,7 +57,7 @@ public:
      * \param bytesPerBlock A vector for each image segment. Each inner vector
      *        contains the compressed size for each block in the segment,
      *        in bytes.
-     * \param compressionRatio Ratio used for J2K compression
+     * \param isNumericallyLossless Flag whether compression was lossless
      * \param numRowsPerBlock The number of rows per block.  Defaults to no
      * blocking.
      * \param numColsPerBlock The number of columns per block.  Defaults to no
@@ -68,7 +68,7 @@ public:
     CompressedSIDDByteProvider(const DerivedData& data,
                                const std::vector<std::string>& schemaPaths,
                                const std::vector<std::vector<size_t> >& bytesPerBlock,
-                               size_t compressionRatio,
+                               bool isNumericallyLossless,
                                size_t numRowsPerBlock = 0,
                                size_t numColsPerBlock = 0,
                                size_t maxProductSize = 0);
--- a/six/modules/c++/six.sidd/include/six/sidd/CompressedSIDDByteProvider.h
+++ b/six/modules/c++/six.sidd/include/six/sidd/CompressedSIDDByteProvider.h
@@ -67,6 +67,7 @@ public:
     CompressedSIDDByteProvider(const DerivedData& data,
                                const std::vector<std::string>& schemaPaths,
                                const std::vector<std::vector<size_t> >& bytesPerBlock,
+                               double compressionLevel = 1,
                                size_t numRowsPerBlock = 0,
                                size_t numColsPerBlock = 0,
                                size_t maxProductSize = 0);
--- a/six/modules/c++/six.sidd/include/six/sidd/CompressedSIDDByteProvider.h
+++ b/six/modules/c++/six.sidd/include/six/sidd/CompressedSIDDByteProvider.h
@@ -57,6 +57,7 @@ public:
      * \param bytesPerBlock A vector for each image segment. Each inner vector
      *        contains the compressed size for each block in the segment,
      *        in bytes.
+     * \param compressionRatio Ratio used for J2K compression
      * \param numRowsPerBlock The number of rows per block.  Defaults to no
      * blocking.
      * \param numColsPerBlock The number of columns per block.  Defaults to no
@@ -67,7 +68,7 @@ public:
     CompressedSIDDByteProvider(const DerivedData& data,
                                const std::vector<std::string>& schemaPaths,
                                const std::vector<std::vector<size_t> >& bytesPerBlock,
-                               double compressionLevel = 1,
+                               size_t compressionRatio,
                                size_t numRowsPerBlock = 0,
                                size_t numColsPerBlock = 0,
                                size_t maxProductSize = 0);
--- a/six/modules/c++/six/source/CompressedByteProvider.cpp
+++ b/six/modules/c++/six/source/CompressedByteProvider.cpp
@@ -25,6 +25,27 @@
 #include <six/ByteProvider.h>
 #include <six/CompressedByteProvider.h>
 
+namespace
+{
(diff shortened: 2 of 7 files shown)


asylvest:
So both 2500C Table A-3 and SIDD Volume 2 Table 2-8 just refer to the BIIF Profile which is available at https://isotc.iso.org/livelink/livelink/fetch/2000/2122/327993/327973/654328/6208440/documents/BPJ2K_01_00.pdf.  According to Table 9-2 there:

```
Set to the approximate number of bits-per-pixel-perband
for the compressed image.

Nxyz = JPEG 2000 numerically lossless, where
"xyz" indicates the expected achieved bit rate (in
bits per pixel per band) for the final layer of each tile.
The decimal point is implicit and assumed to be one
digit from the right (i.e. xy.z).

Vxyz = JPEG 2000 visually lossless, where "xyz" is
the target or expected bit rate (in bits per pixel per
band) for the final layer of each tile. The decimal
point is implicit and assumed to be one digit from
the right (i.e. xy.z).
```

So the current SIX implementation is assuming a) that a numerically lossless compression was performed and b) that what's passed in is bitRate/numBitsPerPixel to make the math work out (the * 10 is for NITF formatting since there's no decimal point) - this was almost certainly just put in really quickly without any thought to correctness as J2K compression was never fully implemented in here.

It seems like what we should do is replace `OPT_J2K_COMPRESSION` with two variables - one for compression type (numerically lossless vs. visually lossless - that way you can populate `N` vs. `V` in `COMRAT`) and a double which is the bit rate in bits/pixel/band (and document these parameters!).  I don't remember if OpenJPEG has an interface to give you an estimated bit rate - I can point you to some code that might populate COMRAT - or if you need to do the math yourself by compressing first.  If your compressed 100x100 pixel mono image was 2000 bytes, your bit rate would be `2000 * 8 / 10000 = 1.6`.

Since the caller has to perform the compression themselves, they should know what the total compressed number of bytes are.  So maybe have SIX's `CompressedByteProvider` take that in at construction instead of the `double` you added?

JonathanMeans:
I think I can just sum over the `bytePerBlock` input vector and work out the bitrate from that. I'm having trouble with numerically vs visually lossless compression, though. I don't see anything in our compressor interface that would set it one way or another. Can I make any assumptions here?

six/modules/c++/six/source/CompressedByteProvider.cpp
```diff
@@ -25,18 +25,48 @@
...
 #include <six/CompressedByteProvider.h>
 
+namespace
+{
+size_t countCompresedBytes(
+        const std::vector<std::vector<size_t> >& bytesPerBlock)
+{
+    size_t sum = 0;
+    for (size_t ii = 0; ii < bytesPerBlock.size(); ++ii)
+    {
```
asylvest: Can store a const ref to `bytesPerBlock[ii]` here

six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -304,8 +305,14 @@ void NITFWriteControl::initialize(mem::SharedPtr<Container> container)
                 subheader.getImageCompression().set("C8");
 
                 //calculate comrat
-                int comratInt = (int)((j2kCompression * nbpp * 10.0) + 0.5);
-                std::string comrat = FmtX("N%03d", comratInt);
+                const int comratInt = (int)((j2kCompression * nbpp * 10.0) + 0.5);
```
asylvest: Use `static_cast` here and below rather than C style casts

six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -304,8 +305,14 @@ void NITFWriteControl::initialize(mem::SharedPtr<Container> container)
...
+                const bool isNumericallyLossless = (bool)mOptions.getParameter(
+                        OPT_J2K_COMPRESSION_LOSSLESS, Parameter(false));
+                // We are assuming the image is not compressed so heavily as
+                // to be visually lossy
+                const char comratChar = isNumericallyLossless ? 'N' : 'V';
+
+                const std::string comrat = FmtX("%c%03d", comratChar, comratInt);
```
asylvest: Avoid using `FmtX()` (under the hood it uses a fixed-length char array.  Use `std::ostringstream` with `std::setw()`.

six/modules/c++/six.sidd/include/six/sidd/CompressedSIDDByteProvider.h
```diff
@@ -57,6 +57,7 @@ class CompressedSIDDByteProvider : public six::CompressedByteProvider
      * \param bytesPerBlock A vector for each image segment. Each inner vector
      *        contains the compressed size for each block in the segment,
      *        in bytes.
+     * \param compressionRatio Ratio used for J2K compression
```
asylvest: Not sure if you had finished the code up and were ready for review or not, but need more comments here to explain what this is a ratio of

JonathanMeans:
Updated.

six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -304,9 +306,18 @@ void NITFWriteControl::initialize(mem::SharedPtr<Container> container)
                 subheader.getImageCompression().set("C8");
 
                 //calculate comrat
-                int comratInt = (int)((j2kCompression * nbpp * 10.0) + 0.5);
-                std::string comrat = FmtX("N%03d", comratInt);
-                subheader.getCompressionRate().set(comrat);
+                const int comratInt = static_cast<int>(
```
asylvest: Use `math::round()` here to be a bit more explicit

six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -304,9 +306,18 @@ void NITFWriteControl::initialize(mem::SharedPtr<Container> container)
...
+                        (j2kCompression * nbpp * 10.0) + 0.5);
+                const bool isNumericallyLossless = static_cast<bool>(
+                        mOptions.getParameter(
+                        OPT_J2K_COMPRESSION_LOSSLESS, Parameter(false)));
+                // We are assuming the image is not compressed so heavily as
+                // to be visually lossy
+                const char comratChar = isNumericallyLossless ? 'N' : 'V';
+                std::stringstream comratStream;
```
asylvest: Use `ostringstream`

asylvest:
This looks good to me.  PR is still marked as WIP - is there anything else you want in here before merging?

JonathanMeans:
Oops! No, that's it.

PR #240, opened 2018-06-25
Fix metadata for test

JonathanMeans:


six/modules/c++/samples/test_create_sidd_with_compressed_byte_provider.cpp
```diff
@@ -1027,8 +1028,25 @@ void writeSIDD(const std::string& filename, bool shouldCompress)
     bytesPerBlock[0].push_back(NITRO_IMAGE.width * NITRO_IMAGE.height * NUM_BANDS);
     std::auto_ptr<six::sidd::DerivedData> data = createData(
             types::RowCol<size_t>(NITRO_IMAGE.height, NITRO_IMAGE.width));
-    six::sidd::CompressedSIDDByteProvider byteProvider(*data, schemaPaths,
-            bytesPerBlock, 1);
+
+    // The ByteProvider can do all the NITFWriteControl setup for you.
```
asylvest: I'm confused... how is this taking a different path through the code?  It looks like the constructor that was called previously was doing fundamentally the same thing?

JonathanMeans: This will call the `initialize()` at CompressedByteProvider.cpp:77, whereas the previously-used ctor will call the `initialize()` at line 52. The latter now sets compression information in the NITF header, which was breaking the test since our data isn't actually compressed. When you pass in a NITFWriteControl to the constructor it assumes you've set up it up exactly how you want it, and doesn't touch it anymore.

asylvest: OK gotcha - thanks

PR #243, opened 2018-08-21
Added missing LHC:RHC type to DualPolarizationType string conversion

JonathanMeans:


asylvest:
Oh, I know why we didn't implement it like this originally... we want these to throw if the value is set to `NOT_SET`.  So, just add in the missing type(s) rather than doing this.

PR #248, opened 2018-10-22
Sync externals

dstarinshak:
Pulling in latest Nitro and CODA-OSS changes


JonathanMeans:
It looks like the CODA-OSS Python bindings are compiled to Python3 rather than Python2. Let me fix this quick first.

PR #255, opened 2018-11-21
Update CSM to version 3.0.3

JonathanMeans:


asylvest:
Let's keep the ability to still build against 3.0.1 for awhile in case we need it for ELT/MSP compatibility.  So, default to 3.0.3 but have a configure option to allow 3.0.1.

JonathanMeans:
Done.

PR #260, opened 2018-12-27
Fix how NITFWriteControl is constructed to work with NITFHeaderCreato…

JonathanMeans:
…r initialization

This fixes the the bug, but I want to propose a couple broader changes to prevent similar bugs. 

We don't need the mutable versions of `getOptions()`. The options are used in initialization, so we shouldn't be able to change them later. Just set it up how you need it before giving it to `NITFWriteControl`.

Instead of having stuff like 

    NITFWriteControl::getOptions() { return mNITFHeaderCreator->getOptions(); }

`NITFWriteControl` should just own everything (so we don't have to worry about keeping the parent class up to date), and `NITFHeaderCreator` should hang on to all its dependencies by a reference or non-owning pointer. 

This last suggestion is a bit of a long-shot, but it seems needlessly complicated to make the caller have to keep track of when and whether to call `initialize()`. It would be nice to just rework all the `WriteControl` stuff to not even have a public `initialize()` method and just remove the constructors that allow it to be in an invalid state afterward. Probably not worth the effort, but I get a bit grumpy whenever I have to work this out again....

(I suppose no one will see this for a couple more days, so I'll start working on a new PR for the first two points...)

### DIFF
--- a/six/modules/c++/six/source/NITFWriteControl.cpp
+++ b/six/modules/c++/six/source/NITFWriteControl.cpp
@@ -43,9 +43,20 @@ NITFWriteControl::NITFWriteControl(mem::SharedPtr<Container> container)
 }
 
 NITFWriteControl::NITFWriteControl(const six::Options& options,
-                                   mem::SharedPtr<Container> container)
+                                   mem::SharedPtr<Container> container,
+                                   const XMLControlRegistry* xmlRegistry)
 {
     mNITFHeaderCreator.reset(new six::NITFHeaderCreator(options, container));
+    if (xmlRegistry)
+    {
+        setXMLControlRegistry(xmlRegistry);
+    }
+}
+
+void NITFWriteControl::setXMLControlRegistry(const XMLControlRegistry* xmlRegistry)
+{
+    mNITFHeaderCreator->setXMLControlRegistry(xmlRegistry);
+    WriteControl::setXMLControlRegistry(xmlRegistry);
 }
 
 void NITFWriteControl::initialize(const six::Options& options,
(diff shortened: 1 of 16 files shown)


six/modules/c++/six/source/NITFWriteControl.cpp
```diff
@@ -43,9 +43,20 @@ NITFWriteControl::NITFWriteControl(mem::SharedPtr<Container> container)
...
 
 NITFWriteControl::NITFWriteControl(const six::Options& options,
-                                   mem::SharedPtr<Container> container)
+                                   mem::SharedPtr<Container> container,
+                                   const XMLControlRegistry* xmlRegistry)
 {
     mNITFHeaderCreator.reset(new six::NITFHeaderCreator(options, container));
+    if (xmlRegistry)
+    {
+        setXMLControlRegistry(xmlRegistry);
```
asylvest: Avoid calling virtual functions in constructors.  This'll actually do what you want in this case, but for code clarity add a private non-virtual `setXMLControlRegistryImpl()`, call that one here, and update `NITFWriteControl::setXMLControlRegistry()` to turn around and call `setXMLControlRegistryImpl()`

asylvest: Except shoot, that turns around and calls `WriteControl::setXMLControlRegistry()` which is virtual too... so I guess add a protected `WriteControl::setXMLControlRegistryImpl()` and have `NITFWriteControl::setXMLControlRegistryImpl()` call that one.

asylvest:
I agree that getting rid of the non-const `getOptions()` makes sense.

six/modules/c++/six/source/NITFWriteControl.cpp
JonathanMeans: I've updated the ctor, and removed the non-const `getOptions()`

asylvest:
Looks good!

PR #265, opened 2019-03-01
Fixed Projection Params

clydestanfield:
outPixelStart and currentOffset were passed in, in the wrong order (see line 226). This swaps them to be correct.

### DIFF
--- a/six/modules/c++/scene/source/ProjectionPolynomialFitter.cpp
+++ b/six/modules/c++/scene/source/ProjectionPolynomialFitter.cpp
@@ -82,8 +82,8 @@ ProjectionPolynomialFitter::ProjectionPolynomialFitter(
              jj < mNumPoints1D;
              ++jj, currentOffset.col += skip.col)
         {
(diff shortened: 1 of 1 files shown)


asylvest:
FYI @joseph-walters @dstarinshak this broke with the VDP updates - you'll want this update before doing any more testing

asylvest:
@JonathanMeans can you in a future PR add a unittest to help catch things like this?

JonathanMeans:
@asylvest Are you just looking for a regression-test sort of thing where we capture the current behavior to catch if it changes? As far as verifying against real data, I'm not sure how to go about that.

asylvest:
@JonathanMeans yeah pick the sample SICD we have on the wiki that has real data on it, save off just the XML and put it in the repo if we don't have it already, then have a test which projects a few points using ProjectionPolynomialFitter.  A regression test is fine, treating the current behavior as correct (but in your local copy, undo the fix from this PR to ensure that the regression test would have caught it).

PR #266, opened 2019-03-05
Add unittest for projection polynomial fitter

JonathanMeans:


JonathanMeans:
I suppose it's slightly redundant with the new Python tests, but... may as well?

PR #272, opened 2019-04-19
Add slantToOutput polynomials to Python layer

JonathanMeans:
This is almost ready to go. There's just a question of how closely points should match after going from slant to output and back to slant. When I run the small SICDs in cropped_nitfs, they're matching exactly. If I run something a bit more full, the result is in the right neighbourhood, but can be off by 5 or so.

### DIFF
--- a/six/modules/python/six.sicd/tests/test_projection_polynomials.py
+++ b/six/modules/python/six.sicd/tests/test_projection_polynomials.py
@@ -24,6 +24,8 @@ if __name__ == '__main__':
                                      'the output plane, and write it as an SIO')
     parser.add_argument('--schema', help='Specifiy a schema or '
                         'directory of schemata')
+    parser.add_argument('--order', type=int, default=3,
+                        help='Order of polynomials to calcualte')
     parser.add_argument('input', help='Input SICD')
     args = parser.parse_args()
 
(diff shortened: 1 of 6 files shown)


six/modules/python/six.sicd/tests/test_projection_polynomials.py
```diff
@@ -0,0 +1,75 @@
...
+    polynomialFitter = SixSicdUtilities.getPolynomialFitter(complexData)
+    sampleSpacing = RowColDouble(complexData.grid.row.sampleSpacing,
+                                 complexData.grid.col.sampleSpacing)
+    offset = RowColSizeT(complexData.imageData.firstRow,
+                         complexData.imageData.firstCol)
+    scpPixel = RowColDouble(complexData.imageData.scpPixel.row,
+                            complexData.imageData.scpPixel.col)
+
+    toSlantRow, toSlantCol = polynomialFitter.fitOutputToSlantPolynomials(
+        offset, scpPixel, scpPixel, sampleSpacing, 9, 9)
```
asylvest: Make this 9 an input argument and default it to 3 (so we don't overfit).  Just one input argument to use for all four spots where there's a 9 is good.

asylvest:
The Python wrapping code looks correct, so I think the fact that the round-tripping isn't perfect is ok

six/modules/python/six.sicd/tests/test_projection_polynomials.py
JonathanMeans: Done

PR #274, opened 2019-05-01
Merge master into sidd 1.1

JonathanMeans:


asylvest:
So... right.  Obviously a zillion things changed and I'm not sure of a good way to tell in the diff what is just master coming in vs. merge conflicts you had to resolve.  I could do a diff between this branch and master to more clearly see the differences but then that would include already-reviewed changes / refactorings before this merge.

Is there anything specific that you're concerned about that you think I should focus on?

JonathanMeans:
The biggest changes I think were pulling GeoInfo and MatchInformation from six/sicd to six

asylvest:
OK so I'm going to do this the easy way I think... accept this PR as-is, then once SIDD 1.1 is ready for prime time, I'll do a review of this branch into master.

PR #276, opened 2019-05-08
Rename crop sicd

JonathanMeans:
This is ugly.... but I think it's the right incantation

### DIFF
--- a/six/modules/python/six.sicd/source/six_sicd.i
+++ b/six/modules/python/six.sicd/source/six_sicd.i
@@ -36,6 +36,7 @@
 #include "six/sicd/AreaPlaneUtility.h"
 #include "six/sicd/GeoLocator.h"
 #include "six/sicd/SICDWriteControl.h"
+#include "six/sicd/Utilities.h"
 #include <numpyutils/numpyutils.h>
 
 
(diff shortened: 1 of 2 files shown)


six/modules/python/six.sicd/source/six_sicd.i
```diff
@@ -123,28 +124,47 @@ Data* readNITF(const std::string& pathname,
...
+%auto_ptr(six::sicd::ComplexData);
+%include "six/sicd/SICDMesh.h"
+%include "six/sicd/ComplexData.h"
+%include "six/sicd/Utilities.h"
+%rename (cropSICD) cropSICDWrap;
+
+%inline %{
+
 
+void cropSICDWrap(const std::string& inPathname,
```
asylvest: I don't quite follow... these are named as `cropSICDWrap()` and then you have a `%rename` to change `cropSICD()` to `cropSICDWrap()`.  Don't we want to go the other direction?  That is, don't we want the function to be called `cropSICD()` when we call in Python?

JonathanMeans: The rename changes `cropSICDWrap` to `cropSICD` for the Python level.

asylvest: Ahhhhh - I get it.  OK cool.

Issue #279, opened 2019-05-28
Chipped Images


johnpdpkarp:
The CSM API is intended to work with full image coordinates.  It appears that, as currently implemented, if presented with an image chip, the SICD sensor model offsets the input pixel coordinate based on the FirstRow and FirstCol as defined in the XML.  This is not correct.  These would be appropriate values to return via the overloaded getImageStart() method.   Tools exploiting CSM are expected to handle the conversion from chipped coordinates to full image coordinates.

asylvest:
@johnpdpkarp you are correct that the current implementation for SICD and SIDD chipped products is that all coordinates are treated as with respect to the chipped product, not the original full image product.  So are you saying that, for example, `groundToImage()` should take in the ECEF location and return the pixel location in full image coordinates (and `getImageStart()` returns nonzero to reflect the chipping)?

johnpdpkarp:
The intent of the CSM API is for the sensor model to operate in "full image" coordinates.  Whatever software tools exploit CSM are supposed to address how full image coordinates relate to the image chip, and imageToGround() should be passing in "full image" coordinates and groundToImage() should be returning "full image" coordinates.  Since the SICD and SIDD sensor models use the entire SICD/SIDD XML structure as their "state", and thus know about the FirstRow and FirstCol values, it would make sense for getImageStart() to return these values.  This method is really just intended to be "informational" and should have no bearing on the results of imageToGround() or groundToImage().

PR #285, opened 2019-08-16
Assume full image coordinates in toPixel/fromPixel

JonathanMeans:


six/projects/csm/include/six/csm/SIXSensorModel.h
```diff
@@ -362,10 +362,11 @@ class SIXSensorModel : public csm::RasterGM
...
     /**
     * Returns the starting coordinate for the imaging operation.
-    *
-    * \return Always returns (0, 0)
+    * In case you have a chipped image, this will return the pixel
+    * offset to the chip.
+    * \return starting coordinate for the imaging operation
     */
-   virtual csm::ImageCoord getImageStart() const;
+   virtual csm::ImageCoord getImageStart() const = 0;
```
asylvest: This was already declared pure virtual in `RasterGM`.  So, it can be simply removed here.

six/projects/csm/source/SICDSensorModel.cpp
```diff
@@ -387,6 +383,12 @@ csm::ImageVector SICDSensorModel::getImageSize() const
     return csm::ImageVector(mData->getNumRows(), mData->getNumCols());
 }
 
+csm::ImageCoord SICDSensorModel::getImageStart() const
+{
+    return csm::ImageCoord(mData->imageData->firstRow,
```
asylvest: Add a comment here and in the SIDD one summarizing what John explained as the convention in #279.  That way when we look back at this in two years we'll be able to convince ourselves we're accounting for AOIs the right way.

PR #291, opened 2019-11-19
Pick up non-SICD/SIDD DES when reading

JonathanMeans:
The fixes the added mesh issue.

### DIFF
--- a/six/modules/c++/six/source/Container.cpp
+++ b/six/modules/c++/six/source/Container.cpp
@@ -103,4 +103,14 @@ Data* Container::getData(const std::string& iid, size_t numImages)
 
     return getData(dataID);
 }
+
+void Container::addDESSource(nitf::SegmentSource source)
+{
+    mDESSources.push_back(source);
+}
+
+const std::vector<nitf::SegmentSource>& Container::getDESSources() const
+{
+    return mDESSources;
+}
 }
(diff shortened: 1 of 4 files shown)


six/modules/c++/six/source/Container.cpp
```diff
@@ -103,4 +103,14 @@ Data* Container::getData(const std::string& iid, size_t numImages)
 
     return getData(dataID);
 }
+
+void Container::addDESSource(nitf::SegmentSource source)
```
clydestanfield: Pass in SegmentSource as const ref.

six/modules/c++/six/source/NITFHeaderCreator.cpp
```diff
@@ -108,7 +108,7 @@ NITFHeaderCreator::NITFHeaderCreator() :
     loadXmlDataContentHandler();
 }
 
-NITFHeaderCreator::NITFHeaderCreator(mem::SharedPtr<Container> container) : 
+NITFHeaderCreator::NITFHeaderCreator(mem::SharedPtr<Container> container) :
```
clydestanfield: Are we using mem::SharedPtr here to match an existing interface? If not we should use std::shared_ptr

JonathanMeans: Basically, yes. There's a whole call hierarchy of `mem::SharedPtr` going on here. I think doing a separate PR that's kind of a general C++11 update pass (or at least just the smart pointers) would make more sense than making piecemeal changes.

PR #292, opened 2019-12-05
Move Classification from CollectionInformation to ComplexData

JonathanMeans:
This should eliminate the need for ComplexCollectionInformation


six/modules/c++/six.sicd/include/six/sicd/ComplexData.h
```diff
@@ -275,16 +276,17 @@ class ComplexData: public Data
...
-     *  Maps to SICD/CollectionInfo/Classification
      */
     virtual const Classification& getClassification() const
     {
-        return collectionInformation->getClassification();
+        mClassification.level = collectionInformation->getClassificationLevel();
+        return mClassification;
     }
 
     virtual Classification& getClassification()
```
vkrishna1997: Why were we overloading this function in the first place?

JonathanMeans: You mean specifically the non-const accessor? Clients can go through it to change security options in the NITF header (see NITFReadControl). I'd rather not expose this and just have a 'setSecurityOption` method instead, but then I'd rather not break more code than I have to.

vkrishna1997:
Looks good to me!

PR #293, opened 2019-12-05
Regenerate Python bindings

JonathanMeans:


vkrishna1997:
Not sure why swig doesn't like this but the changes make sense

PR #294, opened 2019-12-06
CPHD1.0.1

vkrishna1997:


six/modules/c++/cphd/include/cphd/CPHDXMLControl.h
```diff
@@ -45,24 +37,31 @@ namespace cphd
...
     /*!
      *  \func CPHDXMLControl
-     *
      *  \brief Default constructor
+     *
+     *  \param log provide logger object
+     *  \param ownLog flag indicates if logging will be taken care of by user
      */
-    CPHDXMLControl();
+    CPHDXMLControl(logging::Logger* log = NULL, bool ownLog = false);
```
JonathanMeans: Prefer `nullptr` for new code

six/modules/c++/cphd/include/cphd/CPHDXMLControl.h
```diff
@@ -71,7 +70,7 @@ class CPHDXMLControl : public six::XMLParser
      *  Calls toXML
      *  \return XML String
      */
-    std::string toXMLString(
+    virtual std::string toXMLString(
```
JonathanMeans: Why are these virtual? Future-proofing?

Would it make sense to inherit from `six::XMLControl`?

six/modules/c++/cphd/include/cphd/CPHDXMLControl.h
```diff
@@ -45,24 +37,31 @@ namespace cphd
...
+    //! Destructor
+    virtual ~CPHDXMLControl();
+
+    /*
+     *  \func setLogger
+     *  \brief Handles setting logger based on log provided and ownLog flag
      *
-     *  \brief Constructor with custom log option
+     *  \param log provide logger object
+     *  \param ownLog flag indicates if logging will be taken care of by user
```
JonathanMeans: The flag is for whether the object is responsible for deleting the logger on destruction, isn't it?

six/modules/c++/cphd/source/CPHDXMLControl.cpp
```diff
@@ -991,1374 +113,40 @@ std::unique_ptr<Metadata> CPHDXMLControl::fromXML(
...
+    std::unique_ptr<CPHDXMLParser> parser;
+    parser.reset(new CPHDXMLParser(version, false, mLog));
+    return parser;
+}
 
-    // Optional
-    const XMLElem rcvCollectionXML = getOptional(timelineXML,
-                                                 "RcvCollectionStart");
-    if (rcvCollectionXML)
+std::string CPHDXMLControl::getVersionFromURI(const xml::lite::Document* doc)
```
JonathanMeans: This never gets called

six/modules/c++/cphd/include/cphd/Metadata.h
```diff
@@ -71,6 +71,8 @@ struct Metadata : MetadataBase
     size_t getCompressedSignalSize(size_t channel) const override;
     bool isCompressed() const override;
 
+    std::string getUri() const;
```
JonathanMeans: I don't think the uri stuff is necessary here. The only usage I can see is in `CPHDXMLControl::toXMLImpl`. But the function it's calling wants a version, not a uri.

six/modules/python/six/tests/runTests.py
```diff
@@ -70,31 +70,36 @@ def runCPHDTests(cphdDir):
...
-    print('Running test_round_trip')
+        cphdPathname = os.path.join(cphdDir, os.listdir(cphdDir)[i])
 
-    if success != 0:
-        if os.path.exists('output.cphd'):
-            os.remove('output.cphd')
-        print('Error running test_round_trip')
-        return False
+        # Generate CPHD1.0 output file
+        success = subprocess.call([writeCphd,
```
JonathanMeans: If I'm understanding this right, you're pairing each CPHD in the test directory with a schema of the same version, and doing a round-trip/compare.

It's not valid to assume that `os.listdir` will return files in a certain order. I *think* the xercesc/xml.lite layer is smart enough to find the appropriate schema if you give it the directory containing the schemas.

six/modules/c++/cphd/include/cphd/CPHDXMLControl.h
vkrishna1997: Yeah future-proofing was the intent.

So I tried inheriting from XMLControl, but XMLControl assumes the metadata is a six::Data object instead of the metadata object which I'm using.
Other than the validation function, which is static, I wouldn't be reusing/overriding anything in the XMLControl class.

JonathanMeans: Aargh, that's annoying, but yeah makes sense.

six/modules/c++/cphd/include/cphd/CPHDXMLControl.h
vkrishna1997: Yeah that's right. I took a wild guess based on the context it was being used in ...

six/modules/c++/cphd/include/cphd/Metadata.h
vkrishna1997: So I implemented the CPHDXMLParser constructor such that it takes in a URI, as opposed to the sicd parsers which take in a version to build the URI (See `six:ComplexXMLParser` constructor).
Sicd is able to do that because they expect a fixed format for the URI in all the schemas (i.e. urn::SICD:x.x.x). So it just plugs in the appropriate version in the string and pass the schema validation check.
But because the CPHD1.0.1 uri is different, I think the only way to do write out the same uri would be to just store the URI somewhere and use it again.

JonathanMeans: Ok, that makes sense. Just update the relevant parameter names in `CPHDXMLControl` to reflect that.

PR #297, opened 2020-01-08
Initialize default PVP parameters

vkrishna1997:


six/modules/c++/cphd/include/cphd/PVP.h
```diff
@@ -396,22 +415,36 @@ struct Pvp
...
-     *  \brief Validate and set the metadata of the specified parameter
+     *  \param hasAmpSF flag indicates if optional param AmpSF exists
+     *  \param hasFxN1 flag indicates if optional param FxN1 exists
+     *  \param hasFxN2 flag indicates if optional param FxN2 exists
+     *  \param hasToaE1 flag indicates if optional param ToaE1 exists
+     *  \param hasToaE2 flag indicates if optional param ToaE2 exists
+     *  \param hasTDIonoSRP flag indicates if optional param TDIonoSRP exists
+     *  \param hasSignal flag indicates if optional param signal exists
+     */
+    void specifyOptionalParameters(bool hasAmpSF = false, bool hasFxN1 = false, bool hasFxN2 = false,
```
JonathanMeans: Now that I actually see it, I think having a function that takes 7 boolean arguments is going to be hard to read and easy to misuse. Can you find a more explicit interface, a separate function for each parameter or passing flags, or something?

(And yeah, I realize this was my bad idea...)

six/modules/c++/cphd/source/PVP.cpp
```diff
@@ -42,8 +43,25 @@ APVPType::APVPType() :
...
+    initialize();
+    specifyOptionalParameters(hasAmpSF, hasFxN1, hasFxN2,
+                              hasToaE1, hasToaE2, hasTDIonoSRP,
+                              hasSignal);
+}
+
+void Pvp::initialize()
+{
+    // Rest of the parameters are already initialized
+    // in PVPType constructor
```
JonathanMeans: I'd rather have everything specified here, so the information is all in one place in the code.

six/modules/c++/cphd/include/cphd/PVP.h
```diff
@@ -435,6 +501,27 @@ struct Pvp
...
+
+    /*
+     * Set default size and format for each parameter
+     */
+    void setDefaultValues(size_t size, const std::string& format, PVPType& param);
+
+    /*
+     *  Optional parameter flags
+     */
+    bool mAmpSF;
```
JonathanMeans: Can you make it work without this? I'd like checking for optional parameters to be consistent across the library, if possible.

vkrishna1997: CPHD03::VBM.h does this too so I was trying to be consistent with that, although CPHD03 does only has 3 optional parameters. I can move the flags to PvpBlock which is the CHPD10 equivalent of VBM.h if that's slightly better.
Or did you have some other case of optional parameter handling elsewhere in mind?

JonathanMeans: Fair, that makes sense.

six/modules/c++/cphd/source/PVP.cpp
```diff
@@ -31,6 +31,7 @@ namespace cphd
 const size_t PVPType::WORD_BYTE_SIZE = 8;
 
 PVPType::PVPType() :
+    // Overwrite these default values if required
```
JonathanMeans: This comment is stale, right/

PR #307, opened 2020-01-24
Added user version specification

vkrishna1997:
I haven't done much testing yet (some python sicd generation test was failing), but wanted to know if this is basically what you wanted.

### DIFF
--- a/six/modules/c++/cphd/include/cphd/Metadata.h
+++ b/six/modules/c++/cphd/include/cphd/Metadata.h
@@ -58,6 +58,7 @@ namespace cphd
 struct Metadata : MetadataBase
 {
     //! Default constructor
+    //! Initializes CPHD version to default version specified in FileHeader
     Metadata();
 
     /*
--- a/six/modules/c++/cphd/include/cphd/Metadata.h
+++ b/six/modules/c++/cphd/include/cphd/Metadata.h
@@ -58,11 +58,7 @@ namespace cphd
 struct Metadata : MetadataBase
 {
     //! Default constructor
-    Metadata()
-    {
-        // Default version defined in cphd::FileHeader
-        mVersion = FileHeader::DEFAULT_VERSION;
-    }
+    Metadata();
 
     /*
      * Getter functions
@@ -74,10 +70,10 @@ struct Metadata : MetadataBase
     size_t getCompressedSignalSize(size_t channel) const override;
     bool isCompressed() const override;
 
-    //! Get version for new header
+    //! Get CPHD version
     std::string getVersion() const;
 
-    //! Set version for new header
+    //! Set CPHD version
     void setVersion(const std::string& version);
 
     //!  CollectionInfo block.  Contains the general collection information
--- a/six/modules/c++/cphd/include/cphd/Metadata.h
+++ b/six/modules/c++/cphd/include/cphd/Metadata.h
@@ -80,25 +80,6 @@ struct Metadata : MetadataBase
     //! Set version for new header
     void setVersion(const std::string& version);
 
-    /*
-     *  \func getUri
-     *  \brief get URI of XML
-     *
-     *  Get the uri stored in metadata object
-     *  Not intended to be used by user
-     */
-    std::string getUri() const;
-
-    /*
-     *  \func setUri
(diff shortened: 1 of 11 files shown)


six/modules/c++/cphd/source/CPHDReader.cpp
```diff
@@ -68,6 +68,7 @@ void CPHDReader::initialize(std::shared_ptr<io::SeekableInputStream> inStream,
     }
 
     mMetadata = CPHDXMLControl(logger.get(), false).fromXML(xmlParser.getDocument(), schemaPaths);
+    mMetadata->setVersion(mFileHeader.getVersion());
```
JonathanMeans: I would expect the XML parser to be able to completely construct the CPHD Metadata. Also, since the version and URI reflect the same information, I'd like to be able to keep them in sync without having to remember to do it. For example, with SICD the Metadata just stores the version, then uses `versionToURI` and `getVersionfromURI` to talk to the XML layer.

JonathanMeans: But yeah, this is basically what I need!

vkrishna1997: So to my understanding, SICD is able to keep track of the version and URI together because the URI is always in the same format "urn:SICD:<version>".
So it's easy to get the version from the uri or vice versa.
Because the URI's for CPHD are formatted differently for each version, I couldn't think of a way to handle the version and the URI as one (even though it would make sense to do so).

One hacky solution might be to do some regex parsing to get the version  from the URI, but even that requires some assumptions about the format of the URI.

JonathanMeans: Could you just hardcode the mapping somewhere?

vkrishna1997: Also, for that same reason I couldn't depend on XMLParser to get the version from the XML to update the version in the metadata object, so I have to get it from the header. And CPHDReader seemed like the best place to do that, but yeah I can see how ugly that solution looks lol

vkrishna1997: Haha yeah that works. Like a text file in the conf directory or something?

JonathanMeans: It should be fine to do it in the code itself. This shouldn't ever change; they'll just publish a new version if they want to do something different.

six/modules/c++/cphd/include/cphd/Metadata.h
```diff
@@ -59,6 +60,8 @@ struct Metadata : MetadataBase
     //! Default constructor
     Metadata()
     {
+        // Default version defined in cphd::FileHeader
+        mVersion = FileHeader::DEFAULT_VERSION;
```
JonathanMeans: Move the implementation to `Metadata.cpp`

six/modules/c++/cphd/include/cphd/Metadata.h
```diff
@@ -71,8 +74,11 @@ struct Metadata : MetadataBase
     size_t getCompressedSignalSize(size_t channel) const override;
     bool isCompressed() const override;
 
-    std::string getUri() const;
-    void setUri(const std::string& version);
+    //! Get version for new header
```
JonathanMeans: I would remove the second part of the comment, because writing the header isn't the only reason to ask for the version.

six/modules/c++/cphd/source/CPHDXMLControl.cpp
```diff
@@ -99,7 +107,7 @@ std::unique_ptr<xml::lite::Document> CPHDXMLControl::toXML(
 
 std::unique_ptr<xml::lite::Document> CPHDXMLControl::toXMLImpl(const Metadata& metadata)
 {
-    return getParser(metadata.getUri())->toXML(metadata);
+    return getParser(VERSION_URI_MAP.find(metadata.getVersion())->second)->toXML(metadata);
```
JonathanMeans: If user provided an invalid version, then accessing the result of `find` will cause UB.

JonathanMeans:
Is this still WIP?

PR #308, opened 2020-01-28
Pull up DomainType accessors to base cphd Metadata class

JonathanMeans:


six/modules/c++/cphd/include/cphd/Metadata.h
```diff
@@ -71,6 +71,13 @@ struct Metadata : MetadataBase
     size_t getCompressedSignalSize(size_t channel) const override;
     bool isCompressed() const override;
 
+    /*!
+     * Get domain type
+     * FX for frequency domain,
+     * TOA for time-of-arrival domain
+     */
+    cphd::DomainType getDomainType() const override;
```
vkrishna1997: Already in cphd namespace right?
Is the explicit namespace here just style convention?

vkrishna1997:
Other than that one change, looks good to me

PR #309, opened 2020-01-28
Regenerate Python bindings

JonathanMeans:
Forgot to configure with SWIG


vkrishna1997:
Always fun lol. Looks good!

PR #312, opened 2020-02-05
Don't alter previous processing data

JonathanMeans:


vkrishna1997:
Yeah that fix makes sense.

PR #315, opened 2020-02-11
Compile in schema install location so users don't have to remember to…

JonathanMeans:
… deal with it


six/modules/wscript
```diff
@@ -11,6 +11,10 @@ def configure(conf):
     # have to set NITF_PLUGIN_PATH
     conf.env['enable_static_tres'] = True
 
+    conf.define('DEFAULT_SCHEMA_PATH',
+                os.path.join(os.path.abspath(conf.options.prefix),
```
vkrishna1997: Where does conf.options.prefix get defined?
Does the variable have to get updated separately for different modules' schema paths if they wanted to depend on this parameter?

JonathanMeans: It's the value of what you pass to `--prefix=...` (or whatever it defaults to if you don't pass anything). So basically this always points to `<install>/conf/schema/six`.

vkrishna1997: Ah makes sense

PR #321, opened 2020-02-24
Bump version to 3.0.0

JonathanMeans:


vkrishna1997:
Does this effect any version upgrade code?

JonathanMeans:
No, this is the version of the library itself, not for any given standard.

PR #324, opened 2020-03-05
Auto-formatter pass for cleaner pending PR

JonathanMeans:
Blind application of clang-format

### DIFF
--- a/six/modules/c++/cphd/unittests/test_pvp_block_round.cpp
+++ b/six/modules/c++/cphd/unittests/test_pvp_block_round.cpp
@@ -19,33 +19,33 @@
  * see <http://www.gnu.org/licenses/>.
  *
  */
-#include <cstdio>
-#include <stdlib.h>
-#include <iostream>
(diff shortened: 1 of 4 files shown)


six/modules/c++/cphd/unittests/test_pvp_block_round.cpp
```diff
@@ -162,58 +172,79 @@ bool runTest(bool scale, const std::vector<std::complex<T> >& writeData,
 TEST_CASE(testPVPBlockSimple)
 {
     const types::RowCol<size_t> dims(128, 256);
-    const std::vector<std::complex<sys::Int16_T> > writeData =
+    const std::vector<std::complex<sys::Int16_T>> writeData =
```
vkrishna1997: Looks like these are fine in C++11, but do we have to support pre c++11 as well?

JonathanMeans: Not any more.

PR #340, opened 2020-05-27
SIDD 2.0 Updates

vkrishna1997:
Only the files in six/modules/c++ need be reviewed, the rest are swig generated python wrappers


six/modules/c++/six.sidd/source/DerivedXMLParser200.cpp
```diff
@@ -806,14 +820,32 @@ void DerivedXMLParser200::parseColorSpaceTransformFromXML(
 {
     XMLElem manageElem = getFirstAndOnly(colorElem, "ColorManagementModule");
 
-    parseEnum(getFirstAndOnly(manageElem, "RenderingIntent"),
-            transform.colorManagementModule.renderingIntent);
+    std::string renderIntentStr;
```
vkrishna1997, line 819: Had to do this work around because fixing wasn't possible. Either because of invisible chars or #defines with the same name somewhere else. Wasn't able to figure out why.

six/modules/c++/six.sidd/include/six/sidd/DerivedDataBuilder.h
```diff
@@ -80,7 +80,7 @@ class DerivedDataBuilder
      *  \param regionType What kind of region type scheme does the product use
      *  \return Reference to self
      */
-    virtual DerivedDataBuilder& addGeographicAndTargetOld(RegionType regionType);
+    virtual DerivedDataBuilder& addGeographicAndTarget(RegionType regionType);
```
bfisc, line 83: do you currently use:
1. addGeographicAndTarget & addGeoData for SIDD 2.0 and just addGeographicAndTarget for 1.0
2. addGeoData for SIDD 2.0 and addGeographicAndTarget for 1.0

based on the comments looks like the second. but i thought based on the code needed to use both (1.)

vkrishna1997: Nope the main change of this PR was to make it 2.
Should have mentioned that to you lol

bfisc: 👍 thanks, this looks correct then.

six/modules/c++/six.sidd/source/DerivedXMLParser200.cpp
```diff
@@ -543,7 +543,17 @@ void DerivedXMLParser200::parseLookupTableFromXML(
...
+    parseString(getFirstAndOnly(bandElem, "Algorithm"), bandAlgo);
+
+    if (bandAlgo == "1DLUT")
+    {
+        band.algorithm = BandEqualizationAlgorithm("LUT 1D");
+    }
+    else
+    {
+        band.algorithm = BandEqualizationAlgorithm(bandAlgo);
+    }
```
bfisc, line 556: this is nasty. is this because of the change in SIDD spec from `LUT 1D` -> `1DLUT`? should this if be added to `BandEqualizationAlgorithm` instead?

six/modules/c++/six.sidd/source/DerivedXMLParser200.cpp
bfisc: works for me

bfisc: (and by works for me i mean the workaround make sense to me)

six/modules/c++/six.sidd/source/DerivedXMLParser200.cpp
vkrishna1997: So the script that generates the enums from a text file requires that the enum be the same as the string. And I'm like 99% sure c++ doesn't allow you to make an enum starting with a number.
I could just go in and manually change the Enums.h file, but if someone re-generates it again the change will be overwritten. 

The script lives somewhere in Coda so I can maybe add functionality to that to optionally specify a different string for each enum but that seemed like a not super essential update to me atm.

Also, sucks that the enum the schema wants is 1DLUT instead of LUT1D ... hopefully we can just change that for the next version.

So yeah open to second opinion.

bfisc: that is fair. makes sense to me. 

i would say this code block is better than changing schema and/or coda.

six/modules/c++/six.sidd/source/DerivedXMLParser200.cpp
```diff
@@ -1872,34 +1910,37 @@ XMLElem DerivedXMLParser200::convertDigitalElevationDataToXML(
...
 {
-    std::string model;
+    common().parseEarthModelType(getFirstAndOnly(geoDataXML, "EarthModel"),
+                        geoData->earthModel);
 
-    common().parseEarthModelType(getFirstAndOnly(geographicElem, "EarthModel"), geographicAndTarget.earthModel);
-    common().parseFootprint(getFirstAndOnly(geographicElem, "ImageCorners"), "ICP",
-        *geographicAndTarget.imageCorners);
+    common().parseFootprint(getFirstAndOnly(geoDataXML, "ImageCorners"), "ICP",
+                   geoData->imageCorners);
```
bfisc, line 1920: so not to be that person, but what decided the amount of white space for second line? seems like -8 from the open paren on previous line?

bfisc: i should have saved my white space commentary for line 1928... unfortunately i am allowed only 1 white space comment per review...

six/modules/c++/six.sidd/source/GeographicAndTarget.cpp
```diff
@@ -45,10 +45,7 @@ bool GeographicCoverage::operator==(const GeographicCoverage& rhs) const
 bool GeographicAndTarget::operator==(const GeographicAndTarget& rhs) const
 {
     return (geographicCoverage == rhs.geographicCoverage &&
-        targetInformation == rhs.targetInformation && 
-        earthModel == rhs.earthModel &&
-        imageCorners == rhs.imageCorners &&
-        validData == rhs.validData &&
+        targetInformation == rhs.targetInformation &&
```
bfisc, line 48: why?

six/modules/c++/six.sidd/source/DerivedXMLParser200.cpp
vkrishna1997: LOL.
In my defense 1928 isn't my code!
But also not in my defense, shouldn't have blindly copied it from SICD ...

six/modules/c++/six.sidd/source/GeographicAndTarget.cpp
vkrishna1997: All of those were specific to SIDD 2.0.0 only, and are all in GeoDataBase now.
We were using the same class for both versions before because of the schema being weird ...

six/modules/c++/six/source/GeoDataBase.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace six
+{
+// const double GeoDataBase::ECF_THRESHOLD = 1e-2;
+
+bool GeoDataBase::operator==(const GeoDataBase& rhs) const
+{
+    return (earthModel == rhs.earthModel &&
+            imageCorners == rhs.imageCorners && validData == rhs.validData
+            && geoInfos == rhs.geoInfos);
```
bfisc, line 34:
```suggestion
            imageCorners == rhs.imageCorners &&
            validData == rhs.validData &&
            geoInfos == rhs.geoInfos);
```

bfisc: consistency with other code. doesn't really matter to me.

PR #343, opened 2020-06-23
Initial cphd bindings

dstarinshak:
New features:
* Initial cphd-python module that supports metadata generation for CPHDv1.0+. Entry-by-entry population of the `cphd::Metadata` class verified from Python using cphdCustom.xml file in the repo. Convenient manipulation of the `PVPBlock` object using structure numpy arrays is not yet supported.

Relevant bugfixes:
* Default PVP format for "signal" field is fixed. CPHDv1.0.x spec states format must be "I8"; it was "F8" before.

Notes:
* The cphd-c++ module makes use of `std::unordered_map` in places. This is not supported in swig v3.x and lower, particularly with string keys. Swig v4.0+ required to wrap these objects.


six/modules/python/cphd/tests/test_make_cphd.py
```diff
@@ -0,0 +1,738 @@
+#!/usr/bin/env python
+
+#
+# =========================================================================
+# This file is part of cphd-python
+# =========================================================================
+#
+# (C) Copyright 2004 - 2015, MDA Information Systems LLC
```
christianwhite1193, line 8: Old copyright header?

six/modules/python/cphd/tests/test_make_cphd.py
```diff
@@ -0,0 +1,738 @@
...
+
+    try:
+        xml_str = xml_parser.toXMLString(metadata)
+        meta_from_str = xml_parser.fromXMLString(xml_str)
+        xml_str_from_meta_from_str = xml_parser.toXMLString(meta_from_str)
+        # TODO: Verify round-tripped strings are identical. Because of the
+        # use of std::unordered_map this is not guaranteed at the moment
+        print('Test passed')
+    except Exception:
+        print('Test failed')
```
christianwhite1193, line 738: missing a newline if we care about being pep8 compliant

six/modules/python/cphd/wscript
```diff
@@ -0,0 +1,6 @@
+distclean = options = configure = lambda p: None
+
+def build(bld):
+  bld.swigModule(name = 'cphd',
+                 use = 'cphd-c++ cphd-c++ types-python mem-python six.sicd-python six-python sys-python numpyutils-c++',
+                 package = 'pysix')
```
christianwhite1193, line 6: same remark about pep8 compliancy (2 --> 4 spaces, 72 character line limit exceeded on 'use' line, newline needed at eof, additional line needed before 'def build')

six/modules/python/cphd/tests/test_make_cphd.py
dstarinshak: Fixed

six/modules/python/cphd/tests/test_make_cphd.py
dstarinshak: `pycodestyle` complains when I add an extra blank line at the end of file.

six/modules/python/cphd/wscript
dstarinshak: Forgot to run `pycodestyle` on wscripts. It's compliant now

six/modules/python/cphd/tests/test_make_cphd.py
christianwhite1193: my bad, I think the print('Test failed') you have is actually "print('Test failed')\n" if you were to display carriage returns, which is compliant

PR #346, opened 2020-09-04
check if the default schema path exists

christianwhite1193:
XMLControl validation was updated to always load a default schema path defined at config time. The default config path points to the install prefix directory. This causes runtime errors when the absolute path to the schema directory has been changed and no SIX_SCHEMA_PATH has been set (for example, renaming your source directory or creating a delivery).

### DIFF
--- a/six/modules/c++/six/source/XMLControl.cpp
+++ b/six/modules/c++/six/source/XMLControl.cpp
@@ -115,9 +115,9 @@ void XMLControl::validate(const xml::lite::Document* doc,
         // log any error found and throw
         if (!errors.empty())
         {
-            for (size_t i = 0; i < errors.size(); ++i)
+            if (log)
             {
-                if (log)
+                for (size_t i = 0; i < errors.size(); ++i)
                 {
                     log->critical(errors[i].toString());
                 }
--- a/six/modules/c++/six/source/XMLControl.cpp
+++ b/six/modules/c++/six/source/XMLControl.cpp
@@ -71,6 +71,15 @@ void XMLControl::validate(const xml::lite::Document* doc,
     std::vector<std::string> paths(schemaPaths);
     loadSchemaPaths(paths);
 
+    if (schemaPaths.empty() && log)
+    {
+        std::ostringstream oss;
+        oss << "Coudn't validate XML - no schemas paths provided "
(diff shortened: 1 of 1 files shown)


chvink:
Pam D mentioned that setting the environment variable didn't work during testing, do you have any idea why it wouldn't?

six/modules/c++/six/source/XMLControl.cpp
```diff
@@ -44,14 +44,16 @@ void XMLControl::loadSchemaPaths(std::vector<std::string>& schemaPaths)
...
         str::trim(envPath);
         if (!envPath.empty())
         {
             schemaPaths.push_back(envPath);
         }
-        else
+        else if (os.exists(DEFAULT_SCHEMA_PATH))
         {
             schemaPaths.push_back(DEFAULT_SCHEMA_PATH);
         }
```
chvink, line 59: I'd log a warning in validate() when loadSchemaPaths() returns no results and the user didn't provide any of their own. I'm OK that loadSchemaPaths() doesn't try to load non-existing paths.

chvink: Also, it's not portable, but for *nix we could set this up to readlink('/proc/self/exe') to find the location of the current executable, then try a few locations relative to that instead of using a build-time DEFAULT_SCHEMA_PATH. There's probably a windows way to do that too. 

I'm satisfied with what's here (plus the log warning) for the purposes if getting things fixed, though.

christianwhite1193:
@chvink, setting SIX_SCHEMA_PATH should actually resolve the problem - I believe the issue Pam D was facing was related to setting the SICD schema path in a config file rather than setting an environment variable. The SICD path in the config was the absolute path of the schema file - because  validation gets called by several classes that either inherit from or are composed with XMLControl (CPHDXMLControl, ComplexXMLControl etc.), validation succeeded when using the ComplexXMLControl (was given the absolute path) but other's failed (were given an empty schema vector).

six/modules/c++/six/source/XMLControl.cpp
christianwhite1193: Updated to log a warning message - also guarding against a null logger.

PR #351, opened 2020-09-21
Write wideband data with CPHD metadata and add a Python list-of-dicts interface for PVP parameters

elliot-klein:
* Add `cphd::CPHDWriter::writeWideband()` to write a NumPy array of `complex64` wideband data along with CPHD `Metadata` and `PVPBlock` objects
*  Add methods to `cphd::PVPBlock` to convert to/from a list of Python dictionaries of PVP data (a list of channels, each with a dict of PVP parameter names mapping to NumPy arrays)
 * Make `cphd::PVPBlock` used `std::map` to store added PVP parameters instead of `std::unordered_map`. This was needed to support older versions of SWIG.
* Add Python tests for these features and move some common test code into `util.py`. To install `util.py` along with the test folder, use `waf install` (not just `waf install --target=test_.....py`)

## Issues
* `cphd::CPHDWriter::writeWideband()` doesn't support compressed data (and there is currently nothing preventing someone from using it to write compressed data and failing)
*  `cphd::CPHDWriter::writeWideband()` calls `cphd::CPHDWriter::write()`, which was previously `%ignore`d by SWIG, so the Python `CPHDWriter` now has both a `write()` and `writeWideband()` method, and it may be unclear which one to use.
* Using `PVPBlock` list-of-dicts with custom PVP parameters and/or multiple components is not fully supported:
  * Using custom `float` or `complex` PVP parameters requires using `astype('float32')` and `astype('complex64')`, otherwise the values are slightly inconsistent when reading to/from a list-of-dicts
  * Different data types in a single parameter (e.g. `X='F8';Y='I8';`) isn't supported, parameters with multiple components can only have one type (this is only an issue for custom parameters, none of the default parameters currently require this).
  * Custom PVP parameters with multiple components aren't supported
  * Default PVP parameters that are not of size 1 or 3 are not supported (not currently an issue since there are none)
  * If using a string PVP parameter that is longer than the specified size, NumPy will silently truncate it (e.g. a parameter specified as `'S8'` but given `'123456789'` will store `'12345678'`)


six/modules/python/cphd/source/cphd.i
```diff
@@ -193,11 +510,45 @@ using six::Vector3;
...
+ * \param PVPBlock (cphd::PVPBlock)
+ * \param widebandArray
+ *        A single, contiguous NumPy array of all wideband data
+          Multiple data channels should be vstack'd together (first channel first)
+ * \param rows
+ *        Expected number of rows in widebandArray
+ * \param cols
+ *        Expected number of cols in widebandArray
+ */
+void writeWideband(PVPBlock pvpBlock, PyObject* widebandArray, size_t rows, size_t cols)
```
dstarinshak, line 532: I think I see what you were trying to do here, but I don't think it will work in general. I think a cleaner approach would be to circumvent using `CPHDWriter::write()` altogether.

The `write()` routine expects to see a giant concatenated blob of wideband data for all the channels at once. It would be nice if all the per-channel wideband were the same size. Then you could have a single array that's shaped `(num_channels, num_rows, num_cols)`. Getting the underlying array pointer would ensure everything's contiguous in memory.

However, each channel can have its own dimensions. In that case, it seems more natural to have the wideband data expressed as a collection (e.g., list, tuple, etc) of numpy arrays. You can't call the `write()` routine directly because there's no guarantee that the individual numpy arrays are contiguous in memory with each other. We could make them contiguous by doing a copy, but that can get very expensive very fast. Instead, we can modify this extended function to do the following:

```
void writeWideband(PVPBlock pvpBlock, PyObject* widebandArrays)
{
    const size_t numChannels = static_cast<size_t>(<Python_API_for_computing_length>(widebandArrays));
    ASSERT(numChannels == mMetadata.data.getNumChannels());
    writeMetadata(pvpBlock);
    if (mMetadata.data.getNumSupportArrays() != 0)
    {
        if (supportData == nullptr)
        {
            throw except::Exception(Ctxt("SupportData is not provided"));
        }
        writeSupportData(supportData);
    }
    writePVPData(pvpBlock);

    for (size_t ii = 0; ii < numChannels; ++ii)
    {
        const types::RowCol<size_t> dims(
            mMetadata.data.getNumVectors(ii),
            mMetadata.data.getNumSamples(ii)
        );
        const types::RowCol<size_t> arrayDims = numpyutils::getDimensionsRC(widebandArrays[ii]);
        ASSERT(dims == arraysDims);
        auto widebandPtr = numpyutils::getBuffer<std::complex<float>>(widebandArrays[ii]);
        writeCPHDData<std::complex<float>>(widebandPtr, numElements, ii);
    }
}
```

Although this duplicates a lot of `CPHDWriter::write()`, it leaves the Python level looking clean.

I would suggest also adding the following to simplify writing from Python. Hide the `CPHDWriter` object from the Python API:

```
%pythoncode
%{
    def write_cphd(
        pathname,
        metadata,
        pvp_block,
        wideband_arrays,
        schema_paths=None,
        num_threads=None,
        scratch_space=None,
    ):
        """Docstring"""
    if schema_paths is None:
        schema_paths = []  # Not sure if this is the best default or checking the user's env
    if num_threads is None:
        num_threads = 0
    if scratch_space is None:
        scratch_space = 4 * 1024 * 1024

    # Support writing a single channel with a single ndarray
    if isinstance(wideband_arrays, np.ndarray):
        wideband_arrays = [wideband_arrays,]

    writer = CPHDWriter(
        metadata,
        pathname,
        schema_paths,
        num_threads,
        scratch_space,
    )
    writer.writeWideband(pvp_block, wideband_arrays)
%}
```

dstarinshak:
Consider running `pylint` on any (non-SWIG-generated) Python files to check style

six/modules/python/cphd/source/cphd.i
elliot-klein: Thanks @dstarinshak! I updated `writeWideband()` and added the `write_cphd()` method.

How would you recommend handling support arrays? `write_cphd()` currently does not accept any as a parameter (but `writeWideband()` does). `CPHDWriter` has [two `writeSupportData()` methods](https://github.com/ngageoint/six-library/blob/master/six/modules/c%2B%2B/cphd/include/cphd/CPHDWriter.h#L264) -- one that writes support arrays individually given their string ID and a data pointer (without including any padding from the metadata), and another that writes all the support arrays given a pointer to contiguous support array data. Two ideas:
- `write_cphd()` could receive a sequence of NumPy support arrays, make a copy so that the data is contiguous, and pass that to `writeWideband()` (no changes needed to `writeWideband()`)
- `write_cphd()` could receive a Python dictionary mapping string support array IDs to NumPy support array data, and `writeWideband()` could individually write each array by ID (maybe modify that `writeSupportData()` version to include padding)

Thank you!

elliot-klein:
> Consider running `pylint` on any (non-SWIG-generated) Python files to check style

Done, thanks!  All lines are <=100 characters long, let me know if you want me to take that down to 80.

six/modules/python/cphd/source/cphd.i
```diff
@@ -143,6 +143,324 @@ using six::Vector3;
...
+            for paramName, (paramGetter, paramObj) in paramsToCopy.items():
+                paramSize = paramObj.getSize()
+                paramShape = (cphdMetadata.getNumVectors(channel),)
+                if paramSize != 1:
+                    # If data is a vector, add another dimension to the array
+                    paramShape += (paramSize,)
+                paramDtype = self._pvpFormatToNPdtype(paramObj.getFormat())
+                channelPVP[paramName] = numpy.empty(shape=paramShape, dtype=paramDtype)
+
+            # Copy PVP data for this channel by vector
```
dstarinshak, line 373: After looking more closely at this, I think we should strive to remove the loop over vectors. It can get very expensive very quickly. I already have an implementation that does this, but I haven't tested it on data containing `AddedPVP` stuff.

First get a mapping of the (valid) PVP data to their sizes, offsets, and formats (see comment above):
```
pvp_info = ...
```

Since the `PVPBlock` is just a giant binary blob in memory, we need to sort the `pvp_info` by offset:
```
from collections import OrderedDict

offsets = np.asarray([val[1] for val in pvp_info.values()])
offset_order = np.argsort(offsets)
sorted_keys = np.asarray(list(pvp_info.keys()))[offset_order]
sorted_pvp_info = OrderedDict([
    (key, pvp_info[key]) for key in sorted_keys
])
```

Next we need to extract the entire `PVPBlock` binary blob. Thankfully, there's an interface to do that:
```
num_vectors = metadata.getNumVectors(channel)
pvp_size = metadata.pvp.sizeInBytes()

pvp_data = np.array(
    num_vectors * pvp_size,
    dtype=np.bytes_
)

pvp_block.getPVPData(
    channel,
    pvp_data.ctypes.get_data(),
)

pvp_data = pvp_data.reshape(num_vectors, pvp_size)
```

The `pvp_data` array now has shape (num_vectors, pvp_bytes) with serialized byte entries. Getting at the per-entry data amounts to little more than slicing and viewing. For instance, if you know the first PVP entry (i.e. offset=0) is 8-byte floating point (i.e. format='F8'), then `pvp_data[:, :8]` will slice out those first 8 bytes across all vectors. To cast to `np.float64`, you'll have to do a copy since our slice is actually striding in memory:

```
pvp_entry = pvp_data[:, :8].copy().view(np.float64).squeeze()
```

The squeeze at the end ensures (N,1) arrays are returned as simple 1d arrays (N,). Putting this in a big loop give you:

```
pvp_dict = {}
pvp_offset = 0
for name, (offset, size, fmt) in sorted_pvp_info.items():
    # Split parameters into a format dict with keys equal to
    # param suffixes. Single parameters have entry suffix
    # Example: 'F8' --> {'': 'F8'}
    # Example: 'A=F8;B=I4' --> {'A': 'F8', 'B': 'I4'}
    suffix_to_format = _split_multi_format(fmt)

    # If a PVP has multiple parameters but all have the same
    # format, collapse them into a single PVP. This is intended
    # to simplify things like 'txPos', where it's more convenient
    # to have an (N,3)-shape array than three length-N arrays
    # for 'txPosX', 'txPosY', and 'txPosZ' separately.
    _size = 1
    formats = list(suffix_to_format.values())
    if _common_format(formats):
        suffix_to_format = {'': formats[0]}
        _size = size

    for suffix, _fmt in suffix_to_format.items():
        _name = name + suffix
        dtype = _format_to_dtype(_fmt)
        length = dtype.itemsize * _size
        vals = pvp_data[
            pvp_offset:(pvp_offset + length)
        ].copy().view(dtype)
        pvp_dict[_name] = vals.reshape(num_vectors, _size).squeeze()
        pvp_offset += length
```

A few of the details from the method above are:

```
def _common_format(formats):
    return len(set(formats)) == 1

def _split_multi_format(fmt):
    if ';' in fmt:
        result = {}
        for _fmt in fmt.split(';')[:fmt.count(';')]:
            assert '=' in _fmt
            name, __fmt = _fmt.split('=')
            result[name] = __fmt
        return result
    return {'': fmt}
```

six/modules/python/cphd/source/cphd.i
```diff
@@ -193,26 +511,134 @@ using six::Vector3;
...
+%{
+    import os
+    import numpy as np
+
+    def write_cphd(
+        pathname,
+        metadata,
+        pvp_block,
+        wideband_arrays,
+        schema_paths=[],
```
dstarinshak, line 594: Avoid mutable default arguments (i.e. lists, dicts, etc.). Python doesn't make a copy when the function is invoked as C++ would. Some squirrely behavior can result. Instead use `None` and parse what `None` means inside the function body.

six/modules/python/cphd/source/cphd.i
```diff
@@ -193,26 +511,134 @@ using six::Vector3;
...
+}
+
 }
 
+%pythoncode
+%{
+    import os
+    import numpy as np
+
+    def write_cphd(
```
dstarinshak, line 589: I guess for consistency with `read` defined below this method should simply be `write`

dstarinshak:
80 characters is standard. If you can get it there, please do so.

dstarinshak:
@elliot-klein, I understand refactoring the PVP-to-dict logic is a big ask. Do you think it's doable by Friday?

six/modules/python/cphd/source/cphd.i
dstarinshak: I think both approaches for writing support arrays sounds good. The first one sounds like less work. It would only involve adding a support_arrays=None kwarg to write_cphd and parsing out what to do with it inside. As long as you document what form supportArrays should take in the docstring, I think we're good.

elliot-klein: How would you recommend reading the support arrays back in for `test_cphd_write_wideband`? The first method works well for writing, but I haven't tested it since it doesn't look like there are Python bindings yet for reading support arrays.

`CPHDReader` loads a `SupportBlock`, and we can get the support array metadata (string identifiers and dimensions) from `Metadata.data.supportArrayMap`.

`SupportBlock` has [methods to read support arrays by string ID into a pre-allocated `BufferView` or a `ScopedArray`](https://github.com/ngageoint/six-library/blob/master/six/modules/c%2B%2B/cphd/include/cphd/SupportBlock.h#L109)-- based on [`Wideband.read`](https://github.com/ngageoint/six-library/blob/master/six/modules/python/cphd/source/cphd.i#L209) (which calls [(1)](https://github.com/ngageoint/six-library/blob/master/six/modules/python/cphd/source/cphd.i#L149), [(2)](https://github.com/ngageoint/six-library/blob/master/six/modules/c%2B%2B/cphd/include/cphd/Wideband.h#L275), [(3)](https://github.com/ngageoint/six-library/blob/master/six/modules/c%2B%2B/cphd/source/Wideband.cpp#L445)), I think we can cast the `__array_interface__` NumPy data pointer from `long long` to a `void*`, and then to a `sys::ubyte*` to create a `BufferView<sys::ubyte>` for `SupportBlock::read()`. That compiles and runs, but segfaults during `cphd::byteSwap()`.

Thank you!

six/modules/python/cphd/source/cphd.i
elliot-klein: Why does `PVPBlock` set `signal` as a `double` [here](https://github.com/ngageoint/six-library/blob/master/six/modules/c%2B%2B/cphd/include/cphd/PVPBlock.h#L436) and in a few other places?  ('`I8'`= 8-byte signed integer) I think this might be related to an issue I'm having reading `signal` from the binary blob

elliot-klein: Was it because `double`s are 8 bytes and `int`s can be 4? (That's what they are on my machine at least).  Using `signed long long` ([which is guaranteed to be at least 8 bytes](https://en.cppreference.com/w/cpp/language/types)) instead of `double` for `signal` and added `'I8'` parameters fixes the issue I was having reading them from the binary blob.

Similarly, `std::complex<double>` is 16 bytes (8 real, 8 imaginary), which causes issues using it for custom `'CF8'` parameters, which are supposed to be 4 real and 4 imaginary bytes. `std::complex<double>` does work for `'CF16'` custom parameters though.

NumPy dtypes are flexible enough to handle different-sized parameters on the Python side, but I'm not sure what we can do in C++ to guarantee that the data being written is the correct size (and account for different C++ type sizes on different machines).

dstarinshak: There was a bug in the `signal` field initially. I think the author just copy-pasted `double` for the datatype and didn't notice it was different for `signal`. When the datatype was fixed from `F8` to `I8`, I'm not sure it was rippled to all other code (or the existing Python bindings) thoroughly. You're right it should be cast to `int64`

In terms of datatype compatibility between C++ and Python, I believe the following should work:
C++  <-->  Python
`double`  <-->  `np.float64`
`std::complex<float>`  <-->  `np.complex64`
`std::complex<double>`  <-->  `np.complex128`

The `sys` module in CODA-OSS offers abstractions for datatypes in Windows and non-Windows. Check out `Conf.h` if you want to see the explicit breakdown. For instance `sys::Int16_T` is a typedef for `int16_t` on Unix systems and `__int16` on Windows. There are no abstractions for floating point (I believe) because it assumes adherence to IEEE standards for things like `float`, `double`, etc. I think the IEEE standards also apply to `complex` in that the real and imaginary entries are side by side in memory.

Does that help at all?

elliot-klein: Yes, thank you, that helps!

Using `sys::Int64_T` for `signal` works correctly, and using other `sys` types for custom PVP parameters works better as well. In `PVPBlock.cpp`, each potential custom PVP parameter format string now corresponds to one of those `sys` types, rather than e.g. using `int` for all the `'I_'` parameters.

There are still a few issues when reading custom PVP parameters back into NumPy arrays:
 - `'F8'` and `'CF16'` (which both use 8-byte `double`s) are close (<`1e-09`), but not quite exact to what was passed in. I think this might be related to how custom PVP data is stored as `six::Parameter`s, which store data as strings, there might be some loss of precision there.
 - `'U1'` and `'I1'` parameters are both incorrect -- not yet sure why this is. The offset looks correct, and we only read one byte from the binary blob
 - NumPy has no complex integer type for `'CI_'` custom parameters. I will try something like [this](https://stackoverflow.com/a/32877245/3822302), but I'm not sure how it will work transferring to C++.

six/modules/python/cphd/source/cphd.i
elliot-klein: `read()` below becomes `Wideband.read()`; I've been reading wideband data by creating a `CPHDReader` and calling `reader.getWideband().read(...)`. As it is now, `write_cphd()` is a standalone method that creates and runs a `CPHDWriter`. I'm happy to rename `write_cphd()` though if you think it makes more sense

six/modules/python/cphd/source/cphd.i
dstarinshak: I suspect the precision issue is related to what template instantiation of `AddedPVP<T>::getAddedPVP()` you're using. Looking at the top of `cphd.i`, it looks like there are only instantiations for `T` in {`unsigned`, `int`, `float`, `std::complex<int>`, `std::complex<float>`, and `std::string`}. Then, inside `pvp_format_to_added_pvp_methods`, the value of `method_name` is set to one of the explicit instantiations of the `getAddedPVP` getter. The value that `method_name` is set to only depends on the first entry of the format string.

For instance, suppose we're calling it on `"F8"` formatted data. We should be pulling out 8-byte, double-precision data, and we should be using `getAddedPVP<double>`. But instead we use `getAddedPVP<float>`. So we end up downcasting in the conversion and throwing away the extra precision below `O(10^-8)`. The same goes for `"CF16"` since we only provide template instantiation for `std::complex<float>` and not `std::complex<double>`. That probably explains the `10^-9` differences.

This probably also explains why `"U1"` and `"I1"` are incorrect. They're both using the template instantiation for `unsigned` and `int` respectively, both of which are 4 bytes here.

As for `"CI*"` formats, I suggest raising a `NotImplementedError` and calling it a day. There are ways around this using a structured array with a custom `dtype`, but we can always create an issue and circle back to that later. FWIW, I believe this would implement it for `"CI32"`:
```
ci32_dtype = np.dtype([
    ('real', np.int16),
    ('imag', np.int16),
])
data = np.empty(shape, dtype=ci32_dtype)
```

six/modules/python/cphd/source/cphd.i
dstarinshak: Sorry, I lost track of this thread. I suggest simply leaving support arrays out of the bindings for the time being. Is there a clean way to recognize if a CPHD contains support data in the bindings? If so, we could log a warning to the user that "support array parsing it not implemented" and that we're only reading and returning a subset of their CPHD. I would rather not raise a `NotImplementedError` since the remainder of the CPHD should be valid for parsing

six/modules/python/cphd/source/cphd.i
elliot-klein: Yes, that fixes the float precision errors, thank you!

I'm still having issues with `'U1'` and `'I1'`. It looks like `sys::Int8_T` is a typedef for `int8_t`, which in turn is a typedef for `signed char` (which is what SWIG expects `setI1AddedPVP()` to receive). In `from_list_of_dicts()`, we're passing it a Python `int` (converted from a `np.int8`), so I wonder if there might be some hidden type conversion going on here. I tried using `struct.pack('b', param_data)` and converting `param_data` to a `ctypes.c_byte`, but the `setI1AddedPVP()` wouldn't accept either, saying that they aren't `signed char`s.

`NotImplementedError` sounds good for `'CI*'` formats. I'll also raise that for custom PVP parameters with multiple formats. I did try using structured arrays and custom NumPy `dtype`s , and I think we should be able to read them from the binary blob more or less as-is, we just aren't able to set them without accessing the binary blob.

six/modules/python/cphd/source/cphd.i
elliot-klein: We can get the number of support arrays with `metadata.data.getNumSupportArrays()`, but for the moment there's not a great place to check that in the bindings, since we aren't extending `CPHDReader` or `SupportBlock` at all.

It's possible to get a `SupportBlock` object with `CPHDReader::getSupportBlock()`, but then there's no Python interface to get the actual support block data. To write a CPHD, you need to provide the support array data (as well as the `SupportBlock` from the metadata), so I don't think it would be possible for someone to load a CPHD file that has support arrays and then e.g. write it to another file without realizing that the support arrays are missing.

six/modules/python/cphd/source/cphd.i
```diff
@@ -143,6 +144,289 @@ using six::Vector3;
...
+            name[3:].lower(): getattr(pvp_block, name)
+            for name in dir(pvp_block)
+            if name.startswith('set')
+            and name[3:].lower() in default_params_in_use
+        }
+
+        # Populate PVPBlock object from pvp_data
+        for channel_index, channel_data in enumerate(pvp_data):
+            for vector_index in range(metadata.getNumVectors(channel_index)):
+                for name, data in channel_data.items():
```
elliot-klein, line 399: It might be good to check that the list of keys in `channel_data` matches the list of PVP params (the required ones, any optional ones, and any added ones)

six/modules/python/cphd/source/cphd.i
```diff
@@ -143,6 +144,289 @@ using six::Vector3;
...
+        # Methods to check if an optional PVP param is set
+        # Using .lower() since name capitalization is not 100% consistent
+        has_optional_param_methods = {name.lower(): getattr(pvp_block, name)
+                                      for name in dir(pvp_block)
+                                      if name.startswith('has')}
+
+        pvp_info = {}
+
+        # Set required and optional PVP parameters
+        for name in dir(metadata.pvp):
```
elliot-klein, line 370: There's a bit of code that's duplicated here from above. Here, we need to get the list of `PVPBlock` setter methods for all default PVP params (required ones plus any optional parameters that are in use) and ideally build a dict that maps PVP param names to setter methods

elliot-klein (commented):
Sorry, I ran out of time to completely wrap this up. I think this is what remains:
* Fix the `'U1'` and `'I1'` PVP params not being correctly converted from the list-of-dicts
* Decide what to do about support array reading (might be just fine as-is for now). I pushed my WIP changes to the `wip-support-arrays` branch of my fork in case you want to use those as a starting point.
* I flagged some duplicated code for generating the list of PVP params in `from_list_of_dicts()`, you might be able to come up with a cleaner way to do that
* Test string and multi-size custom PVP params
* Make any final style fixes
* Update the SWIG-generated files

Thanks again for all your help with this

PR #482, opened 2021-09-09
Make DirectionParameters::calculateWeightFunction() a public function

mbraswell:
This function is useful to be used in other places, so I made it public here.


JDanielSmith:
Merged change from **develop/jdsmith**; see https://github.com/ngageoint/six-library/pull/483

PR #532, opened 2022-01-19
Fix AMP8I_PHS8I reading

andrew-hardin:
There's a bug in the AMP8I_PHS8I reading where only the first block of the complex data buffer is populated. This PR resolves the issue by seeking the buffer to the correct write position.


JDanielSmith:
I'm merging this into https://github.com/ngageoint/six-library/tree/feature/8AMPI_PHSI and from there to **master**.

Issue #534, opened 2022-01-19
Slow performance reading and writing AMP8I_PHS8I


andrew-hardin:
I have a simple processing chain that looks like this:
```
1. Read AMP8I_PHS8I as std::complex<float>.
2. Modify std::complex<float>.
3. Write std::complex<float> as AMP8I_PHS8I.
```
The bottleneck on this process is reading and writing the pixel content, specifically the methods tied to the conversions between AMP8I_PHS8I and `std::complex<float>`. Is there any interest in future pull requests that improve the performance of those methods?

JDanielSmith:
I've set things up to do all the AMP8I_PHS8I -> `std::complex<float>` conversion at once (instead of pixel-by-pixel) for a particular data-set (at least that's the intention).  From what I can tell, the real bottleneck is the *KDTree* structure I'm using to find a (amp, phs) pair for any given `std::complex<float>`. If you've got a different approach, that would be great! :-)

andrew-hardin:
Your note about the KDTree matches some functions identified by `perf`. There may not be a better alternative, but I can't say for sure because I haven't fully wrapped my head around what's going on yet.

Are there any unmerged changes on a branch that I should build on, or, is everything related already merged with `master`?

JDanielSmith:
The only thing not in **master** is your most recent PR; although I would work in the **feature/8AMPI_PHSI** branch.

andrew-hardin:
Thank you for your guidance on where to start - I'll begin taking a look at optimizations to improve performance.

JDanielSmith:
The first thing to try is enabling the multi-threaded implementation by passing a positive value to `ImageData::to_AMP8I_PHS8I()`; see https://github.com/ngageoint/six-library/blob/master/six/modules/c%2B%2B/six.sicd/source/ImageData.cpp

JDanielSmith:
I've exposed the `cutoff` parameter which controls multi-threading for `to_AMP8I_PHS8I()` through the various layers of code; I've also turned on multi-threading by default.

See the latest changes in the https://github.com/ngageoint/six-library/tree/feature/8AMPI_PHSI branch.

JDanielSmith:
This should be better with multi-threading enabled by default: https://github.com/ngageoint/six-library/pull/536

PR #537, opened 2022-01-30
Faster conversion of std::complex -> 8-bit amplitude and phase (3x speedup)

andrew-hardin:
This change introduces a new acceleration structure that can efficiently map from `std::complex` to 8-bit amplitude and phase.
This is accomplished by exploiting the systematic distribution of the 8-bit amplitude and phase values in the original complex domain.

I measured the performance of `master` and this branch. Specifically, the runtime of `ImageData::to_AMP8I_PHS8I` that's called by the `read_8bit_ampphs_with_table` unit test:
- old: 9132ms
- new: 2884ms

This addresses #534 while keeping the mulit-threading performance improvement added by #536.

### DIFF
--- a/six/modules/c++/six.sicd/CMakeLists.txt
+++ b/six/modules/c++/six.sicd/CMakeLists.txt
@@ -6,6 +6,7 @@ coda_add_module(
         source/AreaPlaneUtility.cpp
         source/ComplexData.cpp
         source/ComplexDataBuilder.cpp
+        source/ComplexToAMP8IPHS8I.cpp
         source/ComplexXMLControl.cpp
         source/ComplexXMLParser.cpp
         source/ComplexXMLParser040.cpp
--- a/six/modules/c++/six.sicd/include/six/sicd/ComplexToAMP8IPHS8I.h
+++ b/six/modules/c++/six.sicd/include/six/sicd/ComplexToAMP8IPHS8I.h
@@ -1,0 +1,45 @@
+#ifndef __SIX_SICD_COMPLEXTOAMP8IPHS8I__
+#define __SIX_SICD_COMPLEXTOAMP8IPHS8I__
+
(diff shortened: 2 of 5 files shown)


JDanielSmith:
Thank you! :-)

I've merge this into the https://github.com/ngageoint/six-library/tree/feature/8AMPI_PHSI branch and I've adjusted the code so that a developer can switch between your new implementation and the existing KDTree (at least for now).

The only thing remaining that I want to do is tweak the unit-tests to verify that the same input generates the same output using both techniques.

If all goes well, I should be able to merge this to **master** by the end of the week.

Thanks again! (And if you feel like explaining in a bit more detail what's going on ... dan <dot> smith <at> maxar <dot> <see><oh><em>).

andrew-hardin:
I can add more detail, but be warned that my powerpoint skills aren't amazing and this is best drawn on a whiteboard or graph paper. The 8-bit amplitude and phase values that we're mapping to have a systematic distribution in the complex domain. Here's what I mean by systematic distribution:
radial_plot

A couple things to point out:
- The graph only shows 16 phase angles and 7 magnitudes per angle. A more realistic plot would have 256 phase angles and 256 magnitudes per angle.
- The `phase_direction[]` array contains a unit vector for each possible phase direction. Just the second vector is labeled in the plot.
- The `phase_delta` is the angular difference between each `phase_direction`.
- The `magnitudes` are the ordered set of possible magnitudes and they're identical for each phase angle.

Given any complex value `V`, we can determine the nearest phase angle by computing `round(phase(V) / phase_delta)`. Determining the nearest magnitude value for `V` is slighly more complex:
dot_product

The complex value `V` is projected onto the nearest `phase_direction` via the dot product. The resulting green point is then what's used to find the nearest magnitude via binary search (`std::lower_bound`).

PR #549, opened 2022-04-14
Faster 8AMPI_PHSI -> complex

andrew-hardin:
The multi-threaded improvements in #536 aren't used by `six::sicd::Utilities::getWidebandData()`. That method loops sequentially through the 8-bit amp phase values and calls `six::sicd::Utilities::from_AMP8I_PHS8I()`, which then calls expensive `sincos()`. The end result is long read times for large images.

This pull request improves performance by reusing the conversion implemented in the `ImageData` class. Reusing that method means that `getWidebandData()` now uses a lookup table and runs across multiple threads (at least when reading `8AMPI_PHSI`).

I tried to minimize the number of changes to make it understandable. Modifications are welcome.


JDanielSmith:
Looks good ... I'll take your word for it as nobody else seems to be using this functionality right now. :-)

PR #755, opened 2025-02-11
Adds a check for c++ 17 or newer in six/include/six/Init.h

whitebrianc:
These minor changes are need to allow us to build our software with c++ 17 that incorporates the six-library. My first thought was to attempt to get the six-library as a whole compiling using c++ 17 but that seems non-trivial so I'm submitting this minor change to allow us to move forward.

The test I performed was to make sure this code still compiles using the current c++ standard and it does.


adam-beauchamp:
Rejecting in favor of a new version that won't touch externals

PR #760, opened 2025-03-03
Reduce scope of build to exclude externals tests

Klamptoo:
With this change, tests and unittests within the externals directories are no longer built, and these unittests are no longer run.  Building and testing of the six directory is unaffected.


adam-beauchamp (approved):
LGTM

PR #763, opened 2025-06-16
Allow other C++ std and latest from coda-oss

holmesv3:
This code should be able to be compiled with c++ 14-20 (maybe higher?), but the way the `CMakeLists.txt` is setup it hard-codes the standard to 14. This allows a dependent project to set the standard to a different version.


adam-beauchamp (approved):
LGTM

PR #772, opened 2026-04-28
PVP byte offset fixes

Klamptoo:
Use PVPArrayByteOffset instead of assuming PVP block are adjacent and in the order of the channels array.  Allows blocks to be re-used if specified by the metadata.


adam-beauchamp (approved):
gtg

